PR #7, opened 2016-07-06
added default method to read single value

ctron:
This change adds a new default method the interface which can be used
to read a single value. It is a convenience method for "readValues"
which requires a list as argument and returns a list as result.


ctron:
I know I did not contact you first about that change. But it is a rather small one, so if you don't like it at all, not much is lost.

kevinherron:
I manually merged this after some formatting tweaks. I need to write a "style" guide for contributions that includes formatting preferences, as well as set up the Maven checkstyle plugin (hopefully).

ctron:
Thanks for merging!

If you like I can try and make an Eclipse code formatter settings file. So that it gets formatted automatically upon save. However it am not sure if this can be mapped 100%. Still I could try and you can decide if it is ok.

But having Eclipse auto format on every save definitely saves some time :wink:

kevinherron:
The most important formatting rule is that all indentation is done with 4 spaces, not tabs. There are some quirks about the import order I don't know if the Eclipse formatter can do or not. When editing existing code I never format the whole file - only the section I'm working on.

PR #10, opened 2016-07-07
Added a default method to write a single value

ctron:
This change adds a new default method the interface which can be used
to write a single value. It is a convenience method for "writeValues"
which requires a list as argument and returns a list as result.

---

This would be the pendant to `readValue`.


kevinherron:
I nearly suggested this in the PR for `readValue` :smile:

kevinherron:
You should be able to rebase or amend your commit and force push to your branch, which will update this PR.

ctron:
Looks like that did work! Sorry for that.

Issue #11, opened 2016-07-07
Server: Unable to write to node of type "Variant"


ctron:
I just tried writing a `String` to a Variable Node of Type `Variant`, which fails with "Bad Type".

Debugging into "AttributeWriter.validateDataType" it comes up with a `DataValue`, which gets unboxed to `Variant` and then to `String`. But then the check tries to assign a String to the target Variant, which fails of course.

I would expect that it is possible to write "anything" to a node of type `Variant`.

kevinherron:
I'll have to look into whether that should be possible or not. You can write "anything" do a node of type `Variant` - you just have to encapsulate that anything in a `Variant` first.

ctron:
That is what should happen .. at least I think it is.

So on the client side I am doing something like `write ( new DataValue ( new Variant ( "String" ) ) )`

On the server side it pops up like in the following screenshot:

![milo1](https://cloud.githubusercontent.com/assets/202474/16656515/4eeedf5c-445f-11e6-8fd4-871bc788eb0c.png)

So the original value is the DataValue, from there the Variant and then the String.

Or do you mean I should put a Variant inside the Variant?

kevinherron:
You can't put a Variant in a Variant - I think you're right about this being a bug.

ctron:
Yes I just saw that:

```java
checkArgument(clazzIsArray || !Variant.class.equals(componentClazz), "Variant cannot contain Variant");
```

ctron:
Since it currently does block me, I can work on a fix if you like.

kevinherron:
Sure, in that block you've breakpointed in, you'll need to add a check to see if expected is `Variant`, and if so, allow the write to proceed.

If you could start a unit test for AttributeWriter that would be good too - you could just test this case for now.

ctron:
Alright then ... yes, you are right. A unit test seems to be a good way to start.

PR #12, opened 2016-07-07
#11: Server: Unable to write to node of type "Variant"

ctron:
This change allows to write any value into a node of the
type Variant.

It also adds a unit test for the AttributeWriter which does test for
this case, tests for a simple failure and for the special case of
ByteString to UByte[].

----

That should do the trick. However I am not entirely sure I got your code style. Please be kind :wink:


kevinherron:
Thank you, merged.

PR #13, opened 2016-07-08
Fix a race condition when creating monitoring items.

kevinherron:
The current API requires MonitoringMode.Sampling be used when creating items in order to allow callers a chance to set consumers and do other bookkeeping before values arrive, and then MonitoringMode.Reporting to be set once ready.

The new method allows callers to provide a callback that will be run with a guarantee that it happens before values can arrive.

### DIFF
--- a/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/api/subscriptions/UaSubscription.java
+++ b/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/api/subscriptions/UaSubscription.java
@@ -15,6 +15,7 @@ package org.eclipse.milo.opcua.sdk.client.api.subscriptions;
 
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.BiConsumer;
 
 import com.google.common.collect.ImmutableList;
 import org.eclipse.milo.opcua.stack.core.types.builtin.StatusCode;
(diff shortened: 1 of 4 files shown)


ctron:
I think this solves the technical problem around that. But I think that the API does not look fancy :wink: You can argue a lot about fancy APIs though....

What do you think of some sort of "SubscriptionBuilder" which uses this method in the backend. So you can do something like:

```java
UaSubscription subscription = ...;
Subscriber s = subscription.createSubscriber ();

s.setTimestampsToReturn (TimestampsToReturn.Both );
s.setDefaultParameters ( new MonitoringParameters ( ... ));
s.setDefaultMonitoringMode ( MonitoringMode.XXX );

for ( int i = 0; i < 10; i++ ) {
   s.requestValue ( new NodeId (...) )
    .valueListener ( this::dumpValue )
    .clientIdGenerator ( ClientIds.assigned ( i ) ) // if set it is always applied last
    .parameters ( new MonitoringParameters (...) ) // overrides
    .monitoringMode ( MonitoringMode.YYY )
    .samplingInterval ( ... );
}

{
   s.request ( new ReadValueId () );
}

{
  Request r = s.request ( new MonitoredItemRequest(...) ); // explicit request
  r.valueListener ( value -> System.out.println (value) );
  r.eventListener ( event -> System.out.println (event) );
}

CompletableFuture<List<UaMonitoredItem>> result = s.commit ();
```

kevinherron:
I think there's room for a higher level API or set of builders or something like that in the future, but it would be weird to just do it for this one API call and not the rest. Right now I'm more interested in just getting all the "foundations" to be solid, and once that's done, we can build fancier and even easier to use APIs on top of those foundations.

ctron:
+1

kevinherron:
I think I may reverse my decision to deprecate the API without the callback - I think the current approach of creating items in MonitoringMode.Sampling and then changing them to MonitoringMode.Reporting when you're ready to receive values is perfectly valid.

ctron:
So does this mean that you will completely skip the approach of setting the value listener during the initialization?

kevinherron:
No, what I meant was I will just remove the `@deprecated` annotations and just allow both approaches.

ctron:
+1

kevinherron:
@ctron any idea what this new dependency-ci check is?

ctron:
Nope. Strange.

Issue #19, opened 2016-07-12
Let the SubscriptionListener provide empty default methods


ctron:
I think it would be nice to either have the `SubscriptionListener` provide empty default methods or to have a `DefaultSubscriptionListener` which implements the interface with empty methods.

This makes is easier when you only need one method to be called.

ctron:
If you like I can make the change and do a PR.

kevinherron:
Yes, go ahead and add empty defaults to `SubscriptionListener`.

`onStatusChanged` and `onSubscriptionTransferFailed` should not be ignored though; maybe we could not add default implementations for these?

ctron:
Sorry for the delay. But here it is.

PR #21, opened 2016-07-14
#19: let the SubscriptionListener interface provide default methods

ctron:
The interface SubscriptionListener now provides empty default methods
so that consumers of the events can choose to only implement methods
which are relevant for them.

----

I added the empty default to all methods since sometimes users might just be interested in "the other" event and then you would still have a few empty declarations required.


kevinherron:
Thank you.

PR #24, opened 2016-07-19
slf4j-simple is just a runtime dependency

ctron:
This change does declare the slf4j-simple artifact as "runtime only"


kevinherron:
Made this change in master, don't worry about fixing this up.

PR #25, opened 2016-07-20
Implement UInt8 type and add test case

ctron:
I did things a bit differently than you ;-) So let's discuss it.

But I also added a unit test.

### DIFF
--- a/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/types/builtin/unsigned/newtypes/OutOfRangeException.java
+++ b/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/types/builtin/unsigned/newtypes/OutOfRangeException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Kevin Herron
+ * Copyright (c) 2016 Jens Reimann
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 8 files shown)


kevinherron:
I think most of these changes are good. I would argue for there being only one constructor rather than multiple constructors. And the addition of a `fromByteBits` static method, analogous to `fromIntBits` and `fromLongBits` in UInt32 and UInt64.

As for the naming, I think there's three approaches:

1) Once we're done implementing the new types, name them the same as the old types. This is favorable because it's backwards compatible. There's already code out there that uses these existing names.

2) Name them the way I did in the initial commit, This closely mimics what the datatypes are named in the UA spec. The exception being `UByte`: in the spec "Byte" in unsigned and "SByte" is signed. I think they screwed this up, but it's far too late for anything to be done.

3) Invent our own naming?

ctron:
For the naming: So I did like the approach of having the number of bits in the type name. This is explicit, no guessing. I know in Java it is defined, but in C you never know :wink: So naming it "uint32_t" is better than assuming that a "unsigned int" is really 32bit.

I do understand the historic situation. But we could simply make classes deriving from the new classes and marking them deprecated. That should work.

Adding a `from*` method is fine. And going with a single constructor is fine as well. All my constructors are private, at least they should be. I would suggest to force people to use "valueOf", since then you always can make optimizations like lookup tables. In the deprecated type you can re-add the constructors of course in order to provide compatibility.

kevinherron:
Meh, let's just break the backwards compatibility. That's one of the reasons it's version `0.x.x` right now. We can name them all `UIntN`. These classes should all be final, so deprecated extensions for backwards compatibility won't work.

ctron:
+1

Didn't think about that.

PR #27, opened 2016-07-21
Added hello world example.

hekonsek:
Hi,

I'd like to add some usage examples to Milo. Currently figuring out how to create basic application with Milo requires digging into test code and removing code you don't need.

I have created a simple Hello World example which demonstrates how to use client and custom namespace implementation in order to read/write node value. That should be a good starter for Milo newbies.

BTW I'd like to add more examples and also work on Milo's documentation.

Cheers!

### DIFF
--- a/examples/helloworld/pom.xml
+++ b/examples/helloworld/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.eclipse.milo</groupId>
+        <artifactId>milo-examples</artifactId>
+        <version>0.1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>examples</artifactId>
+
+    <dependencies>
+        <dependency>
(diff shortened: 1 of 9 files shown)


kevinherron:
That would be great. For examples, you could get pretty far if you brought this project back from the dead and updated it to use Milo: https://github.com/digitalpetri/opc-ua-sdk-examples

I was hoping to use the Wiki as a central location to collect documentation.

kevinherron:
On second thought, that project may not be as useful as I thought. It uses something called `CttNamespace` as a source of nodes to program the client examples against and that no longer exists. We may have to come up with a new example namespace that has a few nodes in it.

hekonsek:
I would propose to:
a) add my example, as it is IMHO something many people discovering the project would be interested in i.e. minimal example of client, server and namespace for read/write purposes
b) analyze opc-ua-sdk-examples and reuse what we can
c) extract more examples from existing unit tests
d) work on the documentation in parallel (GitHub wiki should do the job)

What do you think, Kevin?

kevinherron:
Sounds okay to me. When you're ready for me to merge your examples could you please squash everything into a single commit (if you're planning to add more first) and then make sure your commit message has the "Signed-off-by: Your Name youremail@youreemail.com" footer required by the Eclipse Foundation?

ctron:
And don't forget that when it is more than 1.000 lines of code, you will need a full CQ!

kevinherron:
Try to keep the initial commit under 1000 lines then ;)

hekonsek:
I will sign off this commit. Then would I propose to merge what we already have. I will create more pull requests for the next examples as I work on those. Let's focus on merging and documenting this one for the beginning.

hekonsek:
I screwed something when I tried to sign-off my past commit. I created new cleaned pull request with proper changes - #33 . Can we got it merged?

Issue #26, opened 2016-07-21
Resource Leak: DefaultCertificateValidator cannot be stopped


ctron:
Using the `DefaultCertificateValidator` will fork a thread and create a file system watcher. However both resources can never be disposed.

PR #28, opened 2016-07-21
#26: DefaultCertificateValidator may leak resources

ctron:
This change allows one to properly close the validator, cleaning up the
internal thread and file system watcher.

It also adds a new constructor which allows to construct the instance
providing all necessary directories and allows one to disable writing
the rejected certificates.

A test case was also added to test the shutdown.


kevinherron:
Thank you. I fixed a few typos and renamed the test so it can be auto-associated with `DefaultCertificateValidator` by the IDE. This was a good catch. I wouldn't be surprised if there are other issues lurking when it comes to running the client or server in a "container" that can will be stopped and started.

ctron:
I'll let you know when I find them :wink: 

Thank for your quick response!

Issue #30, opened 2016-07-21
Possible DelegateRegistry race condition?


kevinherron:
Reported and discussed here: https://github.com/digitalpetri/ua-client-sdk/issues/9

amitjoy:
I'm getting the same error:

`org.eclipse.milo.opcua.stack.core.UaSerializationException: no decoder registered for encodingId=NodeId{ns=0, id=449}
	at org.eclipse.milo.opcua.stack.core.serialization.DelegateRegistry$Instance.getDecoder(DelegateRegistry.java:210)
	at org.eclipse.milo.opcua.stack.core.serialization.binary.BinaryDecoder.decodeMessage(BinaryDecoder.java:366)
	at org.eclipse.milo.opcua.stack.client.handlers.UaTcpClientMessageHandler.lambda$onOpenSecureChannel$15(UaTcpClientMessageHandler.java:411)
	at org.eclipse.milo.opcua.stack.core.channel.SerializationQueue.lambda$decode$1(SerializationQueue.java:58)
	at org.eclipse.milo.opcua.stack.core.util.ExecutionQueue$PollAndExecute.run(ExecutionQueue.java:107)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)`

kevinherron:
@amitjoy do you have a repeatable example you can share? I've still never seen this reproduced where it wasn't just a ClassLoader issue

amitjoy:
In OSGi environment, it happens if the specific bundle classloader is not delegated to the stack. I actually forgot that the we had discussed long back on delegating OSGi classloader to the stack so that OSGi can load the relevant encoders and decoders in runtime. Found it again while going though our previous discussion. :)

Henry198:
Hello,

I've this problem with execution (Run As --> Eclipse Application) of my program (that uses some of my Eclipse plugins).

```
[ua-shared-pool-0] ERROR org.eclipse.milo.opcua.stack.client.handlers.UaTcpClientMessageHandler - Error decoding OpenSecureChannelResponse: no decoder registered for encodingId=NodeId{ns=0, id=449}
[...]

PR #31, opened 2016-07-21
[#30] Add synchronization to DelegateRegistry initialization

kevinherron:
Ensure the registry cannot be queried for encoders and decoders until
the initial scan for the built-in types and their corresponding
delegates has finished.


ctron:
I just browsed through the changes. To be honest, I don't understand why you make an `AtomicReference` out of it. Since all read and write operations seems to be in one `synchronized` block a simple static field should be fine.

kevinherron:
You're right - it does not have to be an `AtomicReference`, I just have a habit of trying to make any field I can `final`... in this case I don't think it makes a difference either way.

ctron:
Well I think if you really want to keep it, you could also make  a first "get" outside the sync-lock, if the result is non-null, return directly. Otherwise enter the sync-lock and re-test.

Then all cases after the first one would be lock free. At least you used that AtomicReference a bit :wink:

PR #34, opened 2016-07-23
Refactor server startup and shutdown to be async

kevinherron:


kevinherron:
I did the best I could to keep things clean and simple but "async-ifying" things always adds some complexity.

PR #51, opened 2016-09-17
Implementation for XmlDecoder/XmlEncoder

comtel2000:
add implementation for decodeExpandedNodeId and fix decodeLocalizedText error on empty


comtel2000:
Sorry, just found the offical schema and I've to change the NodeClass values as described here:
https://opcfoundation.org/UA/schemas/1.03/Opc.Ua.Types.xsd

kevinherron:
Looking good so far. I'm not sure how much you're planning to implement so let me know when you're ready to have this merged.

comtel2000:
Kevin, for the first part I'm finished. With a full implementation I still struggle with the stream reader api and all the optional nullable xml fields. Maybe other api like JAXB can do the job with less coding and adequate performance. What do you think?

kevinherron:
I have played around with JAXB a little bit but I'm no expert. It seems like you have to use classes generated from the XSD to get anywhere with it, and we can't do that for Milo because the classes already exist. I could be totally wrong though. Also I think that reading from a stream is the most performant way to decode this stuff and it seems to be what the other stacks do as well.

comtel2000:
It's also possible by add annotations to the existing classes or define a external mapping xml. Maybe the enum can be a problem but let me check that as another branch.

kevinherron:
I'll merge this sometime in the next few days.

PR #53, opened 2016-09-20
New subscription-level data change callbacks

kevinherron:
Adds new callbacks to SubscriptionListener that allow for data and event notifications to be received at the subscription-level rather than per-item, if desired.


kevinherron:
@AriSuutariST want to have a look at this?

AriSuutariST:
Looks good to me. I'll have to try this with real application, today or tomorrow.

kevinherron:
The only change I'm still considering is using a single `ImmutableList` containing a `Tuple2` instead of two lists that are assumed to be of corresponding length and position for the pairs.

AriSuutariST:
Ok, that would solve the only thing that was bothering me, going through two lists was kind of old-school :-)

Issue #46, opened 2016-09-15
BinaryDecoder does not preserve null array values


kevinherron:
Part 6, 5.2.4 states: "An Array of zero length is different from an Array that is null so encoders and decoders shall preserve this distinction."

PR #65, opened 2016-10-17
Fix NPE if clientSoftwareCertificates is null

Pro:
If the client doesn't provide any client software certificate, the list is null and causes a NPE.
Fixed with this commit


kevinherron:
Thanks. I have a feeling there's still more NPEs lurking due to the change made in #46

PR #72, opened 2016-11-01
When releasing shared resources, wait for SHARED_EXECUTOR_SERVICE and…

AriSuutariST:
… EXECUTOR_SERVICE to actually stop. Otherwise systems like Apache Tomcat, which check for thread leaks during application undeployment are unhappy.

Signed-off-by: Ari Suutari <ari.suutari@syncrontech.com>


kevinherron:
Thanks!

PR #79, opened 2016-11-07
Move node interfaces from sdk-core back to sdk-server

kevinherron:
Server and client won't be sharing any node interfaces.

This commit also shuffles the nodes around to different packages to mimic what's happening in the client node refactoring.


ctron:
But this PR doesn't create any duplicate packages in in the past?

kevinherron:
No new/duplicate packages. Just moving stuff from `*.sdk.core` to `*.sdk.server`, from `sdk-core` module to `sdk-server` module.

PR #80, opened 2016-11-08
Remove all usage of jacoco plugin

kevinherron:


kevinherron:
I had some grand plans for this that simply aren't being realized right now. Perhaps in the future we can revisit the idea. As it is now the results are being ignored and it just slows down the build.

PR #89, opened 2016-11-18
LDS and LDS-ME support

Pro:
I implemented the LDS (Local Discovery Service) and LDS-ME (Multicast Extenstion) in this PR according to OPC UA Part 12.
This means that the full set of Discovery Services is supported now.

This PR is not yet ready to merge. Please share your thoughts on improvements or my implementation in general.
It is also not yet tested, so just for general review.


kevinherron:
I'll need some time to familiarize myself with Part 12 again before I can review this in meaningful detail.

After a cursory glance though, a couple things occur to me:

1. These additional discovery services should be opt-in, not enabled by default (when using `OpcUaServer`, anyway).
2. In addition to optionally supporting these services in `OpcUaServer`, a new `OpcUaDiscoveryServer` should probably be introduced. This could be the basis that a standalone discovery server would be built on. We could then implement a simple `ExampleDiscoveryServer` using this in the milo-examples module.
3. We should figure out some way to make the dependency on jmdns optional as well, some kind of soft dependency mechanism... I'd like to avoid including this dependency unless you're opting into these services.
4. Because of the bureaucracy involved in adding a dependency to the project, especially at this phase as we prepare for a 0.1.0 release, I think this would be best merged post-0.1.0.

Pro:
Thanks for the feedback. Here some comments:

1. I would suggest adding a boolean to `OpcUaServerConfig` to enable/disable the new discovery services. Good to go?
2. What would be the functionality of the `OpcUaDiscoveryServer`? Only the minimum set of services required for LDS?
- AttributeServiceSet
- MethodServiceSet
- SessionServiceSet
- DiscoveryServiceSet
3. Agree
4. Also agree

kevinherron:
> I would suggest adding a boolean to OpcUaServerConfig to enable/disable the new discovery services. Good to go?

That might work. After thinking about it more it might help if we enumerated the possible scenarios for running a server and possibly discovery server and figure out how the library could best handle it. I've never given much thought to it before - right now I run all my servers on a custom port and just have the server itself implement FindServers and GetEndpoints. But there's obviously more scenarios than that, and I think would be great to support them with the help of this PR.

Without an LDS you have just one basic scenario:
- Running a server without the usage of an LDS at all (just FindServers and GetEndpoints implemented by the server directly). How it works now.

With an LDS the options get more complicated.
- Are we starting the LDS or is it one that will already be present on the system?
- If we will be the LDS, do we simply implement those additional services on the same stack server and port that the server runs on, or do we start up another stack server running on the proper port for discovery (4840)?
- What about registration with the LDS?

I'm sure I'm missing some stuff here.

> What would be the functionality of the OpcUaDiscoveryServer? Only the minimum set of services required for LDS?

Isn't it just the DiscoveryServiceSet? Or is there something in the extended discovery services or profile that would require session/attribute/method?

Pro:
> With an LDS the options get more complicated.

I think we should keep it very simple:
The developer can/should decide (with the boolean config variable) if the current server supports the DiscoveryServiceSet or not. This answers all the questions:

> Are we starting the LDS or is it one that will already be present on the system?

The developer knows that and decides

> If we will be the LDS, do we simply implement those additional services on the same stack server and port that the server runs on, or do we start up another stack server running on the proper port for discovery (4840)?

A LDS may at the same time be a "normal" OPC UA server, so we should not start any additional server automatically. E.g. the developer may also decide that the LDS server is running on port 14840, which is also a valid setup.

> What about registration with the LDS?

There will be an additional config variable which enables or disables the registration with a specific LDS server.

This is also more or less the implementation how we did it in open62541: https://github.com/open62541/open62541/pull/732

kevinherron:
Ok all sounds reasonable so far 👍 

At some point you might rebase or cherry pick from master, I made the existing FindServers implementation a little better in https://github.com/eclipse/milo/commit/492580186fb730c516fb83f56a4be61ecbe204d1

Pro:
I already merged master, see https://github.com/eclipse/milo/pull/89/commits/8eb1926f26d189cddea04b65771ac457e3a32c81

I'm currently implementing the stuff we discussed, and will also try to add examples and unit tests.

kevinherron:
Heads up: sometime this weekend I'm being forced to rewrite history in master. Once that happens you'll need to fetch and reset master in your fork and then rebase this branch onto the "new" master. Make sure you don't merge or the rewritten history may be re-introduced.

Pro:
@kevinherron may I ask why you need to rewrite history? That's really bad practice and will break not only my branch but all the other forks. So there should be a really GOOD reason why you want to rewrite the master...

kevinherron:
Eclipse Foundation is forcing me to do it. I've already lost the appeal. Believe me, this is not my choice.

Legal bureaucracy... 😞

kevinherron:
As you can see by the collateral damage to this PR, the rewrite has taken place. You'll need to rebase your changes rewrite the history of your fork and branch now. Really sorry for the inconvenience.

Pro:
@kevinherron I now implemented Examples for the discovery and cleaned up the code.

Can you review it and tell me if there are any changes required?

Unfortunately I didn't find any way to make the dependency on jmdns optional. There are optional dependencies in Maven, but then you also need to exclude the .java file from compilation. If you do that, the other depending class does not find the excluded class.

I also did not create a separate `OpcUaDiscoveryServer` class, because in my opinion it just makes the code more complicated.

Let me know if you have any suggestions for improvement!

kevinherron:
I'll take a closer look this weekend, but some superficial stuff I see:

- Continuation indents should be 4 spaces if your editor is configurable in that manner. Checkstyle isn't looking for this right now, so that's my fault really.
- I'd suggest the new configuration fields be named `discoveryServerEnabled` and `multicastEnabled`, leading to getters/setters named `isDiscoveryServerEnabled()`, `isMulticastEnabled()`, `setDiscoveryServerEnabled()`, `setMulticastEnabled()`.

Pro:
Done as requested :+1:

kevinherron:
Just wanted to let you know I haven't forgotten about this! I'm just a little pre-occupied trying to get a bunch of work on DataTypes and encoding done for 0.2.0. I'll be sure to merge this for 0.2.0 as well.

opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/services/DiscoveryServices.java
```diff
@@ -0,0 +1,345 @@
...
+
+
+    private Consumer<RegisteredServer> registerServerConsumer = null;
+
+    public DiscoveryServices(UaTcpStackServer server, boolean multicastEnabled) {
+        super(server);
+        registeredServers = new LinkedList<>();
+        registeredServerLastSeen = new HashMap<>();
+        this.multicastEnabled = multicastEnabled;
+        new Timer().scheduleAtFixedRate(new TimerTask() {
```
kevinherron: Use `Stack.sharedScheduledExecutor()` here instead of a new Timer instance? Not sure how this timer instance could be shutdown and cleaned up when the server and stack get shutdown otherwise.

opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/services/DiscoveryServices.java
```diff
@@ -0,0 +1,345 @@
...
+
+            Set<String> hostnames = new HashSet<>();
+            hostnames.add(config.getHostname());
+
+            for (String bindAddress : config.getBindAddresses()) {
+                hostnames.addAll(config.getHostnameResolver().apply(bindAddress));
+            }
+
+            mdnsHelper = new MdnsHelper(hostnames);
+            Thread mdnsThread = new Thread(mdnsHelper);
```
kevinherron: Same concern, it looks like `MdnsHelper` sleeps forever, how do we clean this up?

opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/services/DiscoveryServices.java
```diff
@@ -0,0 +1,345 @@
...
+
+        // set discovery configuration if it is in request (only RegisterServer2)
+        MdnsDiscoveryConfiguration discoveryConfiguration = null;
+        String mdnsName = null;
+        if (requestDiscoveryConfiguration != null) {
+            configurationResults.clear();
+            configurationResults.ensureCapacity(requestDiscoveryConfiguration.length);
+            for (ExtensionObject e : requestDiscoveryConfiguration) {
+                if (discoveryConfiguration == null &&
+                    e.getEncodingTypeId().equals(Identifiers.MdnsDiscoveryConfiguration)) {
```
kevinherron: Does this actually work? Has anything tested code path? 

I would have expected that the encoding id would have to be one of `Identifiers.MdnsDiscoveryConfiguration_Encoding_DefaultBinary` or `Identifiers.MdnsDiscoveryConfiguration_Encoding_DefaultXml` (also accessible as `MdnsDiscoveryConfiguration.BinaryEncodingId` and `MdnsDiscoveryConfiguration.XmlEncodingId`)

Pro:
I don't mind if you change things directly here in this PR (you can push to it).
Otherwise, maybe end of next week I'll have some time to address your comments

Pro:
Just as an additional note: I already tested this Milo LDS implementation in combination with the one I wrote for open62451 and they were able to talk to each other, anyways there may be still some minor issues in the code.

opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/services/DiscoveryServices.java
Pro: Ok, will change that, seems definitely better to use the shared executor.

opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/services/DiscoveryServices.java
Pro: The MdnsHelper thread will stop as soon as the Java VM shuts down. Using the SharedExecutor, it creates a daemon thread. If there are only daemon threads left those will be killed. So it should be fine?

opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/services/DiscoveryServices.java
Pro: Yes, right, it should compare the Encoding ID. I'll change that

Pro:
@kevinherron I changed the things you mentioned in your review

Pro:
@kevinherron I now also added unit tests for discovery.
Is there anything else I can improve?

kevinherron:
The only thing I remember seeing was the issue of whether or not it can shutdown and startup again cleanly within a continually running JVM/env (no standalone threads etc...). I'll gladly take a look at that myself once its merged though.

I'm probably going to create a `dev/0.2.0` branch this week... if the GitHub UI makes it easy to change your base branch to that once I've created it that would be the last thing to do. Then I'll squash+merge this as well as the work I've been doing on the type system in the `type-dictionaries` branch and get things moving...

kevinherron:
Oh, also Travis CI seems to be unhappy about something...

kevinherron:
Branch `dev/0.2.x` now exists. I'd like to start collecting work for 0.2.0 release there.

Pro:
I rebased this PR onto dev/0.2.x

The current error in travis is due to the fact that the identity checking test actually never worked before.
The KeyStoreLoader always initialized the server certificates with `null` because your example-certs.pfx did not contain any certificate with identity `server-ai`.

I changed it to use a valid certificate (https://github.com/eclipse/milo/pull/89/commits/081ea7b6563e70d55e4b9c998f214411682511ef#diff-3ea5a948dc47d2129f93c4929c15fd02R45) and now the identity is actually checked, but can not be validated. Maybe you can check what's wrong here, since I don't have enough insight into the encryption implementation of Milo.

opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/Stack.java
```diff
@@ -105,7 +120,7 @@ public Thread newThread(@Nonnull Runnable r) {
                 }
             };
 
-            SCHEDULED_EXECUTOR_SERVICE = Executors.newSingleThreadScheduledExecutor(threadFactory);
+            SCHEDULED_EXECUTOR_SERVICE = Executors.newScheduledThreadPool(5);
```
kevinherron: Any reason you chose 5 instead of something like `Runtime.getRuntime().availableProcessors()`?

And can you use the overload that also accepts the `threadFactory`

Pro: Done

kevinherron:
I've got this merged into a local copy and am actually looking through the changes in detail in my IDE.

I think aside from modifying it to not use that thread for `MdnsHelper`, it's mostly nitpicky things left. I can make the changes before merging, or I can fork a copy of your repo, make them, and submit them as PRs to your branch. Let me know which approach you'd prefer.

Pro:
Maybe it's best if you just commit the changes into this PR's branch, so everything is kept together.
See https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/

And then merge it. That's probably better for later reference.

opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/services/helpers/MdnsHelper.java
```diff
@@ -0,0 +1,284 @@
...
+        for (InetAddress mdnsIa : addressList) {
+
+            discoveryUrl = "opc.tcp://" + mdnsIa.getHostAddress() + ":" + serviceInfo.getPort() +
+                serviceInfo.getPropertyString("path");
+
+            if (discoveryUrl.endsWith("/")) {
+                discoveryUrl = discoveryUrl.substring(0, discoveryUrl.length() - 1);
+            }
+
+            break;
```
kevinherron: What's the intention here? This break statement makes the loop pointless, only the first element in `addressList` gets used.

Pro: Oh, yeah, this was there for debugging purposes. Should be removed!

kevinherron: Okay, so without the break, it just ends up setting `discoveryUrl` using the last element in `addressList`

Pro: Right... This should return all the possible Discovery Urls for the ServiceInfo and then processRecord should use the Array instead of a single URL

kevinherron:
Assuming the changes I just made to `MdnsHelper` are what you had in mind originally, the last thing left for me to do is figure out how to give MdnsHelper a lifecycle rather than have it be a thread that lives forever.

Pro:
I changed the MdnsHelper to not start its own thread, as you suggested. Fortunately it was quite easy.
I'm not sure why I added this additional thread in the first place.

kevinherron:
I have a bunch of changes for this sitting on my home computer. I'll try to push them Sunday evening or Monday.

kevinherron:
Ugh, pushed the WIP but didn't have my email configured properly on this computer.

Pro:
Your changes look good! I also like it better that way.
Currently I'm not working on this branch/PR so feel free to push/force push if needed.

Pro:
@kevinherron any update on this PR?

kevinherron:
I need to find some time to look at this again. It wasn't finished when I stopped working on it is all I remember. 

I'd still like to merge it into 0.2.0, but making sure all the datatype changes are going to work going forward has taken much longer than expected.

ashhassan100:
> @kevinherron any update on this PR?

Hi @Pro. I have been trying to contact you on your personal website about LDS OPC UA, but the contact form does not work so I thought I would give it a shot here. So essentially, I am working on a LDS implementation and had some questions. I was wondering if you could potentially explain what your process was and some explanation to your implementation. If you could, I would really appreciate it!

Issue #54, opened 2016-09-22
WeakReference<AttributeObserver>>


ergouser:
Is there some reason for using a WeakReference for AttributeObservers? I can think of a number of use-cases where there's a one-to-one relationship between a node and an AttributeObserver and no reason (other than the WeakReference) for any other instance to hold a reference.

kevinherron:
It may just be an artifact of a previous design in the server. Older
versions toyed with a reference counting system for UaNodes and the idea
that nodes kept references to other nodes, and it may have just been that I
didn't want AttributeObserver references that were never cleaned up to leak
when orphaned nodes were still hanging onto references (or something like
that).

It doesn't appear at first glance that they are necessary any more.

It's hard to tell because the design and API in the server SDK is still
either in development or in flux, depending where you look.

ergouser:
Vaguely related (or possibly just my ignorance of usage) is that there does not seem to be a way to change the value of a UaVariableNode node at the server without also triggering fireAttributeChanged.  If you call "setValue" from the server side (upon reading a value from whatever is the real source of the data) an attributeChanged is generated in the same fashion as if a Client had changed the value - not always a desirable side-effect.  The "builder" pattern makes it more difficult to simply subclass the node and add a set method that doesn't trigger the notification.

kevinherron:
Actually my original intent for the `AttributeObserver` was for server-initiated attribute changes. That it happens also when a value is set due to a write is coincidence.

In the "real world" systems I've built on this so far, most writes to real data sources, like PLCs, RTUs, etc... are handled in the write call on `Namespace`, not with observers hanging off each node.

ergouser:
Interesting...  However, for better or for worse I think that OPC is still mostly about tag/value pairs which lends itself well to the UaNode/Observer model.  For example, even though we automatically aggregate PLC register read blocks, they are presented to the user as a single register-tag/value instance/entity.

Put another way, the observer pattern (or a delegate or a subclass) allows the entire read/write operation to be encapsulated in the node structure.  Without that, you have to have some additional mapping (from nodes to read/write entities of arrays of values or similar) in the NameSpace.

With the same philosophy, it might be desirable to pass the MethodInvocationHandler into a UaMethodNode.

These are relatively minor tweaks -  please don't take them as criticisms.  So far everything seems solid and well built.

kevinherron:
I think removing the `WeakReference` is a pretty easy sell. 

[...]

PR #90, opened 2016-11-20
Implement AttributeDelegate concept

kevinherron:
Implements the idea of an `AttributeDelegate` that can be used to hook into the get/set of any attribute without requiring subclassing the `UaNode` implementations.


kevinherron:
Something I find particular powerful about this idea is how they can be chained with `AttributeDelegateChain#create`, allowing you to combine delegates with separate concerns into a single delegate somewhat elegantly. Imagine one kind of delegate that knows how to handle checking if the user is authenticated, another that knows how to read the Description attribute from somewhere, if it exists, another that handles actually doing something with the get/set Value call, etc... You can combine all of these easily so that one node has all this behavior. There is a basic example of this idea in practice in `ExampleAddressSpace`.

Some bike-sheddy stuff under consideration:

1) Should this concept be called `AttributeProxy` instead?
2) On the `AttributeDelegate` interface there is a lack of symmetry between `AttributeDelegate#getAttribute`, which does not throw, and `AttributeDelegate#setAttribute`, which throws `UaException`. This could be remedied by making `setAttribute` return StatusCode instead...

AriZuu:
I browsed through the changes (quite a lot of new code). This looks good!

I was wondering how server works when using something like ExampleNameSpace.addDynamicNodes() when the node is monitored & part of subscription ? Does the framework poll delegate->getValue to find out if value has changed and needs to be pushed to client ?

RestrictedAccessDelegate looks handy, just something I was looking for.

I already implemented most of my UA server based on old examples & discussions, but these changes would give more sophisticated tools for environments where the value actually already "lives" in somewhere (for me values are in memory-based time-series database).

kevinherron:
Yeah, the idea was to simplify some of the programming you have to do when you're actually dealing with a bunch of `UaNode` instances rather than doing things a little bit more dynamic like you mentioned in your case or if you were modeling a PLC where you could do the lookup/browse on the fly. It was originally inspired by the discussion in #54 where the complain was basically that he wanted to subclass nodes to provide functionality like this but the library is not very friendly to subclassing.

Subscriptions still only have one very naive implementation (`SubscriptionModel`) that implements its polling by just calling `NamespaceManager#read`. I fully expect server implementations to provide their own, better, mechanism for updating subscription items at the rate requested by the client. There's too many approaches and too many data source scenarios to build something more generally useful, not to mention that in some applications this subscription mechanism may be part of a product's "secret sauce" to some degree.

AriSuutariST:
Does RestrictedAccessDelegate have effect on item browsing ? If item doesn't  have any read/write access for given identity, the client won't be able to see it's existence at all ?

kevinherron:
> Does RestrictedAccessDelegate have effect on item browsing ? If item doesn't have any read/write access for given identity, the client won't be able to see it's existence at all ?

No, no effect. Clients can see nodes but not read or write the value, depending on the `AccessLevel`.

Browsing is an interesting case as the specification provides no official facility or attributes for restricting what nodes a client sees in a browse, it's basically up to the server to implement it how it pleases. I'm not even sure if the browsing API is extensible enough right now to allow you to implement that or not. It's something that's been on the back of my mind to look into before releasing...

AriSuutariST:
Browse filtering can be done by server-specific implementation of namespace.browse() ?

kevinherron:
> Browse filtering can be done by server-specific implementation of namespace.browse() ?

In theory, yes. In practice... there is an obscene amount of intricate and crucial to be correct if you want to pass UA certification functionality being implemented for the user in `BrowseHelper`. There should be some way for the Namespace implementation to provide some kind of filter that gets used by `BrowseHelper` to narrow down the references it returns... or something.

AriSuutariST:
I kind of noticed that modifying browse() is not easy. My first attempt already failed miserably :-(

Some kind of simple interface that a node could provide for browsing stuff to handle filtering would be nice. But not mandatory, since the users without access cannot read/write anyway. But they will get sad, mumbling "oh, there is that fancy node there, I see it, but I cannot read it :-("

kevinherron:
Hmm... might not be as bad as I thought. All the browsing functionality really gets funneled through `Namespace#getReferences(NodeId)` which would provide a very easy place to install or implement some filtering to the list of references returned. I think `getReferences` just needs to be refactored to pass a little more information... like an `Optional<Session>` that can be used to look up the user if it's a client-initiated browse.

AriSuutariST:
Sounds good. I browsed through the code, looks like getReferences() is not called from many places, so this wouldn't hard to change ?

kevinherron:
> Sounds good. I browsed through the code, looks like getReferences() is not called from many places, so this wouldn't hard to change ?

Seems not. I'll handle this in another PR before release.

PR #95, opened 2016-11-24
Make com.sun JMX handling optional

ctron:
This change differs between plain java, JMX and com.sun JMX in a way
that com.sun JMX is optional and will not fail.

When running inside OSGi it may be that "com.sun" is not available and
thus may fail with a NoClassDefFoundException.


kevinherron:
We're already doing the right thing in the bundle plugin, yeah? Something about setting `com.sun` and friends to optional?

ctron:
Yes, the OSGi bundle was accepting this as "optional" ... however the code did not :wink:

PR #99, opened 2016-12-02
Replace findbugs jsr305 annotations with jetbrains annotations

kevinherron:
It seems we can't track down the author or get licensing approval for the findbugs jsr305 annotation implementation, so replace all these annotations with their equivalent from org.jetbrains:annotations, which is Apache licensed.


ctron:
+1

ctron:
Can you open a CQ for this.

kevinherron:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=12346

Let's give them a day to decide if findbugs jsr305 can be "works with" or not? Makes no difference to me.

PR #110, opened 2017-03-17
Add (some) support for historyRead and historyUpdate operations.

AriSuutariST:
This adds functionality that is required to get the historyRead request to arrive at namespace implementation. It works for me, but I'm not happy about the default historyRead implementation
in AtttributeManager, which just returns StatusCode.BAD for all requested items. I guess it should be polished to return Bad_NotSupported at operation level (but that wouldn't be perfect either, because there might be one namespace that supports history and another that doesn't).

Signed-off-by: Ari Suutari <ari.suutari@syncrontech.com>

### DIFF
--- a/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/OpcUaServer.java
+++ b/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/OpcUaServer.java
@@ -46,6 +46,7 @@ import org.eclipse.milo.opcua.stack.core.BuiltinReferenceType;
 import org.eclipse.milo.opcua.stack.core.ReferenceType;
 import org.eclipse.milo.opcua.stack.core.Stack;
 import org.eclipse.milo.opcua.stack.core.application.UaStackServer;
+import org.eclipse.milo.opcua.stack.core.application.services.AttributeHistoryServiceSet;
 import org.eclipse.milo.opcua.stack.core.application.services.AttributeServiceSet;
 import org.eclipse.milo.opcua.stack.core.application.services.MethodServiceSet;
 import org.eclipse.milo.opcua.stack.core.application.services.MonitoredItemServiceSet;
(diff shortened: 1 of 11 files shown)


AriSuutariST:
Should the ServerNode class have readAttributeHistory which would eventually be delegated to AttributeReader.readAttributeHiistory ? Or separate AttributeHistoryReader class ?

Somehow these methods should also end up calling piece of code from my namespace that provides actual history.

kevinherron:
I don't think `ServerNode` needs additional methods for reading history. They're probably too complicated with all the options and I think they'd realistically need to return `CompletableFuture` because history lookups could be lengthy and it would be different than all the current methods.

kevinherron:
I had some idea for splitting the history methods out of `AttributeServiceSet` into a separate `AttributeHistoryServiceSet` as well as separate `AttributeServices` and `AttributeHistoryServices` implementations but I can't remember exactly why right now...

Ideally if a server doesn't intend to support history, which I think will be the common case, a `Bad_ServiceUnsupported` code is still returned.

AriSuutariST:
I'll try to refactor existing code to AttributeHistoryServiceSet & AttributeHistoryServices to see how it works. This needs more stuff, at least for Session & SessionManager so they have access to those.

AriSuutariST:
How should I handle the difference with server supporting history and one that doesn't ? Depending on that Session should have AttributeHistoryServices implementation that really does something or just dummy, returning Bad_ServiceUnsupported. Currently all those XxxxxServices are created in Session constructor.

kevinherron:
My first thought was in `UaTcpStackServer` just add the dummy default method `AttributeHistoryServiceSet` handler, and if someone using the server wanted a better one they could install the `AttributeHistoryServices` handler themselves.  `UaStackServer` would need another override of `addServiceSet` for the new history services interface.

kevinherron:
Nevermind, `OpcUaServer` replaces those service sets with the session manager in its constructor...

kevinherron:
Maybe just do the separation of services and interfaces, but forget about the Bad_ServiceUnsupported for now. Your operation level StatusCode in the default implementation could be Bad_NotSupported though.

kevinherron:
I think this looks good 👍

AriSuutariST:
I added also similar support for history updates (as I will need that soon myself).

kevinherron:
Can you rebase+squash this into one commit and make sure that resulting commit has your Signed-off-by line at the bottom?

AriSuutariST:
Fixed copyrights & squashed.

kevinherron:
Thanks!

PR #126, opened 2017-07-19
Fix a possible NPE triggered by auto-unboxing

ctron:


kevinherron:
Can't be null except by programmer error, closing.

PR #135, opened 2017-09-07
Fix handling of colons in qualified names

ctron:
I think there is a bug in the qualified name parsing. It looks to me as if only the first colon should be special. However I didn't check that in the spec. @kevinherron you might know this.


kevinherron:
Nice catch. Travis CI has been sporadically killing the build during tests lately 🤦‍♂️

PR #214, opened 2018-01-30
Server and client standalone example

mbruns42:
This example makes it possible to generate jars encapsulating a server or a client built with milo.

It includes README.md files and the build configuration. It also includes a tutorial how to generate openssl certificates without the CertificateBuilder.


kevinherron:
Thank you. I will take some time to review this week. In the interim, you'll need to get an ECA on file before this can be merged. 

https://dev.eclipse.org/eclipse-webhook/services/status_details.php?id=5a7089ad228bb

PR #221, opened 2018-02-02
Squash commits for docker example

mbruns42:
Last version, I promise. Single signed commit. 

Hope you can figure out from the Readme how to use it and from the other version what I actually changed.


kevinherron:
Yikes, something's wrong with this one: 1,170 files changed

mbruns42:
Yes, the problem is, if I squash it, I also squashed all your commits that happened in between.

kevinherron:
I think you may need to rebase before squashing. I might be able to massage the previous PR into shape as long as you checked the box that allows me to push changes when you submitted it.

PR #222, opened 2018-02-02
Add secure standalone client and server example

mbruns42:
Signed-off-by: Melanie Bruns <mbruns42@mail.uni-paderborn.de>


kevinherron:
Now that I've moved the standalone examples to their own module the README needs to explain how the server can generate a keystore as well... 

Are you set on using openssl for the example or should we modify one or both to use the `SelfSignedCertificateBuilder`?

mbruns42:
I modified the server to use the SelfSignedCertificateBuilder and mentioned explicitely in the README that there are two options and which part of the code demonstrates what.

mbruns42:
@kevinherron Do you have any more remarks?

kevinherron:
No, sorry, I just lost of track of this 👍  Thanks for the reminder.

PR #228, opened 2018-02-23
Add example for server with a user database and client connecting to it

mbruns42:
This is an example of a server checking a database of users and hashed passwords for granting access. Should I add the database for an example, too?


kevinherron:
> Should I add the database for an example, too?

Since this is an example, and it's using SQLite?, it would be nice if the database got created and populated when the example started up and noticed it was missing.

upohl:
Now we have a trusted folded and a rejected folder for user databases. This should be similar to the certification folders. The examples creates a databases if no databases exists. The server checks if an user is stored within the trusted database and if the password is correct. If an unknown user tries to connect the username and the hashed password are added to  the rejected database. Afterwards, it is easy to copy or to merge the rejected database into to trusted database to add new users.

kevinherron:
When I run the example I get this in the server logs:

```
05:40:12.887 [ua-shared-pool-1] INFO  o.e.m.e.s.ExampleServerWithPasswordDatabase - db dir: /var/folders/z5/n2r_tpbn5wd_2kf6jh5kn9_40000gn/T/security/db
05:40:13.003 [ua-shared-pool-1] INFO  o.e.m.e.s.ExampleServerWithPasswordDatabase - Connected to user database
05:40:13.006 [ua-shared-pool-1] INFO  o.e.m.e.s.ExampleServerWithPasswordDatabase - Successfully created database:SQLite JDBC
05:40:13.010 [ua-shared-pool-1] INFO  o.e.m.e.s.ExampleServerWithPasswordDatabase - Successfully created table
05:40:13.010 [ua-shared-pool-1] INFO  o.e.m.e.s.ExampleServerWithPasswordDatabase - Connected to user database
05:40:13.014 [ua-shared-pool-1] INFO  o.e.m.e.s.ExampleServerWithPasswordDatabase - SQL Statement: SELECT Password FROM Users WHERE Username=?
05:40:13.016 [ua-shared-pool-1] ERROR o.e.m.e.s.ExampleServerWithPasswordDatabase - Problem accessing user database
java.sql.SQLException: ResultSet closed
	at org.sqlite.core.CoreResultSet.checkOpen(CoreResultSet.java:69)
	at org.sqlite.jdbc3.JDBC3ResultSet.findColumn(JDBC3ResultSet.java:38)
	at org.sqlite.jdbc3.JDBC3ResultSet.getString(JDBC3ResultSet.java:443)
	at org.eclipse.milo.examples.server.ExampleServerWithPasswordDatabase.lambda$new$1(ExampleServerWithPasswordDatabase.java:256)
	at org.eclipse.milo.opcua.sdk.server.identity.UsernameIdentityValidator.validateUserNameIdentityToken(UsernameIdentityValidator.java:125)
	at org.eclipse.milo.opcua.sdk.server.identity.UsernameIdentityValidator.validateUsernameToken(UsernameIdentityValidator.java:68)
	at org.eclipse.milo.opcua.sdk.server.identity.AbstractIdentityValidator.validateIdentityToken(AbstractIdentityValidator.java:54)
	at org.eclipse.milo.opcua.sdk.server.SessionManager.validateIdentityToken(SessionManager.java:543)
	at org.eclipse.milo.opcua.sdk.server.SessionManager.onActivateSession(SessionManager.java:478)
	at org.eclipse.milo.opcua.stack.server.tcp.UaTcpStackServer.receiveRequest(UaTcpStackServer.java:240)
	at org.eclipse.milo.opcua.stack.server.handlers.UaTcpServerSymmetricHandler$2.lambda$onMessageDecoded$0(UaTcpServerSymmetricHandler.java:280)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

Is that expected?

upohl:
That was not expected and occurs when the created database is empty. Should be fixed now.

kevinherron:
I think it would be nice if two things changed:

1. The default username/password for the client example is one that would also work when connecting to the regular example server (i.e. user/password1 or admin/password2).
2. When the server example starts and creates the database, it populates its table(s) with one or both of these users. This way somebody trying to run the example doesn't have to muck around with doing a DB insert to create a user first.

upohl:
Hi Kevin, the problem, which I have with such a solution, is that such "hardcoded" default users are often a security issue or developers forget to remove them from the code. Currently, clients try to connect and all these rejected users are stored in the database db/rejectedUsers.db. Afterwards, it is easy to copy that file manually into the db/trusted folder. As a result, these users are accepted. I think this solution is quite easy and comparable to how certificates are accepted. Our solution enables to train the OPC UA server with new users via "rejected" login attempts. Do you think that adding a better Readme description would help or should I implement your succession?

kevinherron:
Some documentation would help, but imagine this scenario:

* User connects, is rejected
* Copy DB file from rejected to trusted folder to allow
* User 2 connects, is rejected. We don't want to allow User 2 at all.
* User 3 connects, is rejected. Do we copy the file? Would that allow User 2?

upohl:
ok, I see the problems and started to add more operations to interact with the databases to support your usecases. does this fits your requirements? than I would cleanup the code and add javadoc documentation.

kevinherron:
Yes it's fine. I don't think you need to get too carried away because this is just an example.

I don't think anybody will use the "trusted/rejected" paradigm for user management in a real application. This is also why I don't see much risk involved with the example being set up to work "out of the box".

upohl:
I agree that it should not become to complicated, but I'd like to use the example in an opc ua security training rather than in a productive environment.
The example should show the following use cases: 
* How to use username and password login in combination with a local database. 
* How to handle passwords securely 
* The danger of SQL injections in combination with OPC UA.

kevinherron:
Can you guys take care of the IP validation failures?

mbruns42:
We'll take a look at it.

ctron:
> We'll take a look at it.

If you need any help, please let me know.

mbruns42:
Thank you, @ctron.
After taking a closer look, I think this needs some reorganisation, some details should not be here, e.g., changing the log level or manipulating both the "classic" and the standalone examples.

PR #256, opened 2018-03-23
User database example squashed 

mbruns42:
Same intention as in #228 
The examples were moved to the standalone examples and this does not change debug levels anymore or other changes that were unrelated to the actual examples.


mbruns42:
@kevinherron @ctron Here's the cleaned up version of the previous pull request. The pipeline fails with a timeout, somewhere we didn't touch, could you restart it and check if that was just a timing issue? :)

kevinherron:
Stupid tests... I've looked at these and can't figure out why the fail. They *only* fail on Travis CI!

mbruns42:
I had the same issue before. We ended up mocking system time, which also made the tests faster since we didn't have any `Thread.sleep`. You're already using mockito in other parts, maybe that might be an option for you? 

Anyway, any comments on this Pull Request?

kevinherron:
I'm going to check it out and merge it today. Meant to do that after my previous comment but things have been crazy lately 😮

Issue #260, opened 2018-05-03
Provide a way to override endpoint information


ctron:
Following up from the gitter conversion with @mbruns42 

In some scenarios the server reports back with a hostname which the client doesn't understand. Assume you have an OPC UA server running inside a docker container. Connecting to "host-1" and fetching endpoint information might return endpoints with the name "container-1", which the client cannot use. So the client has to override the provided information with the original hostname it used ("host-1") in order to connect.

Currently I do this in `camel-milo` like this: https://github.com/apache/camel/blob/master/components/camel-milo/src/main/java/org/apache/camel/component/milo/client/internal/SubscriptionManager.java#L576-L587

However I do think that there should be a simpler way to achieve the same result. Some out-of-the-box support for overriding various parts of the endpoint information.

kevinherron:
There is already configuration for this: https://github.com/eclipse/milo/blob/b4237b40d81de977ddc1d79f15590d172d9fcbeb/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/api/config/OpcUaServerConfig.java#L69-L77


The effective list of `EndpointDescription` is the product of bind addresses and endpoint addresses (and security profiles).

The ExampleServer populates the endpoint list like this:
```java
        List<String> endpointAddresses = newArrayList();
        endpointAddresses.add(HostnameUtil.getHostname());
        endpointAddresses.addAll(HostnameUtil.getHostnames("0.0.0.0"));
```

but you're free to configure it however you want.

ctron:
Well I was more talking of the client side. In case the server does report the wrong information, or maybe it is not Milo at all :wink: … then you would need to have client side support for this.

kevinherron:
Just to give you an idea of how your application might handle this configuration, what ours does is: Endpoint Addresses is String setting, expecting a comma-separated list of hostnames to use, where a hostname surrounded by `<>` (so... `<myhost>`) is passed to `HostnameUtil.getHostnames()`, the `<hostname>` is a special case for `HostnameUtil.getHostname(), and anything else is used as it's received.

Given that `config.endpointAddressList` is a `List<String>`, then:
```kotlin
val endpointAddresses = config.endpointAddressList.flatMap {
    if ("<.+>".toRegex().matches(it)) {
        val hostname = it.removeSurrounding("<", ">")

        if (hostname == "hostname") {
            setOf(HostnameUtil.getHostname())
        } else {
[...]

PR #265, opened 2018-05-07
[#260] implement EndpointUtil#replaceUrlHostname()

kevinherron:


kevinherron:
Given the number of fix-ups I've made to this I think it's clear another set of eyes is needed... @ctron

ctron:
What if I would need to replace the port in addition?

kevinherron:
Okay, now you can update just the hostname, just the port, or both.

PR #267, opened 2018-05-11
DoS Mitigations

kevinherron:
Implements some basic DoS mitigations:

1) Rate limit inbound connection attempts on a per-remote-address basis. This does not include connections from localhost.

2) Check the message length indicated in the header against max chunk size ASAP. This prevents a maliciously crafted length and buffer from growing beyond max chunk size, as opposed to letting it grow and catching it when the chunk is handled.

3) Limit the number of total connections and connections per remote address.

4) Implement a deadline for receiving the UA TCP Hello message after a connection is established.


kevinherron:
Still needs an overall limit on the number of connections allowed from any given host...

PR #271, opened 2018-05-16
Verify the configured endpoint is present in the CreateSessionResponse

kevinherron:


kevinherron:
At present this has two issues that need to be resolved:

1) if the endpoint URL in the configured `EndpointDescription` was replaced in order to connect there will be no matching endpoints in the list from the `CreateSessionResponse`.
2) the spec actually says client shall verify the *list* returned, not just that a matching endpoint is present in  the list. This implies the client should be caching the list from discovery in order to verify...

PR #293, opened 2018-06-21
Update readme with correct names of profiles and info about user database

mbruns42:
Saw the discussion on gitter about the wrong names of the maven profiles, updates these and added more info about the two examples and their differences.


kevinherron:
Thanks!

PR #295, opened 2018-07-19
Ignore duplicate namespace URIs

jlnr:
Hey and thanks for this wonderful library!

We are using Milo 0.2.1 to connect to a commercial OPC UA product, and ran into an issue where said product returns several namespaces with the URI "null". This then leads to an IllegalArgumentException in Milo because the same value ("null") is inserted into the namespace BiMap twice. And this exception then prevents Milo from learning about all namespaces in the array after the second "null" value.

I am not sure what the best way to handle this (presumably buggy?) server behavior would be. Attached is the simple fix we are currently evaluating internally.

### DIFF
--- a/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/OpcUaClient.java
+++ b/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/OpcUaClient.java
@@ -208,6 +208,11 @@ public class OpcUaClient implements UaClient {
                         uriTable.clear();
 
                         for (int i = 0; i < uris.length; i++) {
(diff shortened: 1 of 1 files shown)


jlnr:
For comparison, the UaExpert GUI shows these "null" namespaces as blank entries, but is still able to work with the server. That doesn't help much because these two namespaces don't contain anything anyway.

screen shot 2018-07-19 at 20 49 47

kevinherron:
Yeah, I'd definitely classify this as buggy server behavior.

I think ignoring duplicates is an appropriate workaround, and perhaps also ignoring null URIs as well. I'll try and see what other stacks are doing in this case. I vaguely remember somebody from Prosys talking about this at the interop last year.

kevinherron:
Will fix in another PR

PR #323, opened 2018-09-24
Minor syntax fix

marcusschneider:


kevinherron:
So... as absurd as it sounds, I can't accept this PR unless you have an Eclipse Contributor Agreement signed and on file for the email address in the commit.

If you don't want to deal with that, I can hijack these changes and commit them myself since they're so minor, but the changes won't be attributed to you.

Let me know what you'd like to do. If you think you might contribute to this or any other Eclipse project in the future it may be worth the time to sign the ECA.

marcusschneider:
Hi Kevin, I fixed that right now.

kevinherron:
Thanks!

PR #349, opened 2018-10-26
use getSubjectAltNameField method and URI constant

abirke:
The method to extract a SAN of a certain type from
a certificate was already implemented but not used
in SessionManager.validateApplicationUri.

Signed-off-by: André Birke <abirke@campus.uni-paderborn.de>


kevinherron:
Thanks for submitting this PR.

It seems there's even more you can do to remove duplication. Remove the method entirely and replace it with a call to `CertificateValidationUtil.validateApplicationUri(...)`: 

https://github.com/eclipse/milo/blob/21557708b4367970aae716a2031af1a73c3c8502/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/util/CertificateValidationUtil.java#L213-L224

abirke:
Cool. I didn't see that

PR #377, opened 2018-12-11
AccessLevel fix

hjparker:
Needed this update to be able to correctly set the access-level for object members. See https://stackoverflow.com/questions/53716525/eclipse-milo-writing-to-an-object-member-variable


kevinherron:
Thanks. Looks like I missed this in refactored NodeFactory in the `dev/0.3` branch as well.

All contributions to an Eclipse project must come from a user with an Eclipse Contributor Agreement on file (not my rules...). 

It's pretty easy to do - just make an Eclipse account and fill out the form.

If you don't want to do this I can re-submit your changes on my own but you would lose authorship. Let me know what you'd like to do.

hjparker:
It is okay, please go ahead and re-submit the changes.

Thanks.

kevinherron:
Okay, the changes are available on the `master` branch now. Thanks again.

PR #379, opened 2018-12-19
Configurable LocaleIds when activateSession

igarciasz:
Currently, session's preferred localeIds are set to none (empty array). With this modification they default to none, but they can be configured by calling OpcUaClientConfigBuilder.setSessionLocaleIds()

Signed-off-by: igarciasz <ignacio.garcia@tecnalia.com>

### DIFF
--- a/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/api/config/OpcUaClientConfig.java
+++ b/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/api/config/OpcUaClientConfig.java
@@ -60,6 +60,11 @@ public interface OpcUaClientConfig extends UaTcpStackClientConfig {
     BsdParser getBsdParser();
 
     /**
+     * @return the list of locale ids in priority order for localized strings
+     */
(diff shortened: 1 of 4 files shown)


kevinherron:
Thanks for your contribution.

Can you make sure to add this new setting the copy tests in `OpcUaClientConfigTest` as well?

igarciasz:
Setting added to the copy tests in OpcUaClientConfigTest

kevinherron:
Thank you. In order to accept any contribution to an Eclipse project it must come from somebody who has a CLA on file. You can make a free Eclipse account and then sign one from there.

igarciasz:
I had already signed the CLA, but it seems there's a problem when the check tries to validate it. Eclipse Foundation also reports a problem with my CLA, maybe they are related:
There is a valid ECA on file for ignacio.garcia@tecnalia.com but we found a problem. This user is missing from the ECA LDAP group. If the problem persists after 10 minutes, please contact webmaster@eclipse.org. (eca-missing-ldap-05)

kevinherron:
Hmm, you're right. I might have made a typo when I put your email address into the tool.

Thanks!

PR #381, opened 2018-12-19
Fix eclipse checkstyle errors

asashour:
I believe there is "maven" checking of checkstyle, but by default eclipse import of the projects, the below cause errors. I believe they are not harmful, and we can add checkstyle checks in Eclipse to match maven, at a later point.

Signed-off-by: Ahmed Ashour <asashour@yahoo.com>

### DIFF
--- a/milo-examples/client-examples/src/main/java/org/eclipse/milo/examples/client/ProsysEventSubscriptionExample.java
+++ b/milo-examples/client-examples/src/main/java/org/eclipse/milo/examples/client/ProsysEventSubscriptionExample.java
@@ -13,7 +13,6 @@
 
 package org.eclipse.milo.examples.client;
 
-
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


kevinherron:
Thank you. In order to accept any contribution to an Eclipse project it must come from somebody who has a CLA on file. You can make a free Eclipse account and then sign one from there.

asashour:
Thanks for reply, it is already done, before submitting, I guess the ip-validation have errors.

image

kevinherron:
Yep, I can see that you've signed in using the committer tool. There seems to be some internal error preventing this integration from working. The online tool has this to say:

> There is a valid ECA on file for asashour@yahoo.com but we found a problem. This user is missing from the ECA LDAP group. If the problem persists after 10 minutes, please contact webmaster@eclipse.org. (eca-missing-ldap-05)

I'll go ahead and merge your changes though, because it does say your ECA is valid.

PR #382, opened 2018-12-21
DateTime to support Instant

### LINKED ISSUE #138: DateTime should provide conversion to Instant
`DateTime` currently provides a conversion to Java `Date`. But it does not provide a conversion to `Instant`, which would also support nano-seconds, which `Date` does not.

asashour:
Fixes #138 

Signed-off-by: Ahmed Ashour <asashour@yahoo.com>


opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/types/builtin/DateTime.java
```diff
@@ -29,7 +30,7 @@
     private final long utcTime;
 
     public DateTime() {
-        this(javaToUtc(System.currentTimeMillis()));
+        this(Instant.now());
```
kevinherron: I wonder if the original default constructor is faster. `Instant.now()` seems to do a lot.

asashour: Pros for Instant: it reserves the nano seconds, while old constructor uses only millis.

Cons: this might be a change in expectations.

No issue, I will revert it.

kevinherron: What if we revert the constructor but provide another static function `DateTime.instant()` that gives you easy access to nano-precision DateTime?

kevinherron: Making the constructor private would need to happen on `dev/0.3` branch. I'm okay breaking APIs since the library is version `0.x` but I try to keep there breaks to the intermediate version numbers.

PR #452, opened 2019-04-09
Documentation

joesan:
Added README to run client examples!

### DIFF
--- a/milo-examples/client-examples/README.md
+++ b/milo-examples/client-examples/README.md
@@ -1,0 +1,9 @@
+# How to use the client examples
+
+To run the client examples., perform the following steps:
+
+1. Import the project into your favorite IDE (IntelliJ or Eclipse)
+
+2. Run any one of the provided implementations of the ClientExample interface
+
+Note that by running the client examples., it will launch a server that runs locally and the client will connect to this local server. So to run the client examples, you do not need to run a server beforehand!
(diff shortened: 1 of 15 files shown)


joesan:
I have no idea on how to get past the buerocracy... it is beurocrazy! My last commit message looks like this:

```
commit 4bdc2c7315b9559c9b04c74d344b824ab07889c0 (HEAD -> documentation, origin/documentation)
Author: Joesan <codeintheopen@gmail.com>
Date:   Tue Apr 9 21:37:04 2019 +0200

    Update

    Signed-off-by: Joesan <codeintheopen@gmail.com>
```
Note sure what else is expected!

kevinherron:
If you squash these all into one commit that has the correct author and footer it should work.

joesan:
Should the Author name match with the Eclipse user name?

kevinherron:
I don't know if the name matters or just the email address.

joesan:
Everything looks good with the Author and Footer, but still I could not make it!

joesan:
Here is the latest commit: 

https://api.github.com/repos/eclipse/milo/commits/54b76ca1d5585c203d235110b307d036ee31690c

Do you see anything wrong?

kevinherron:
The 2nd commit of the 6 doesn't have a signed-off-by in it. Again, squash these into 1 commit and it will be easier.

joesan:
I probably need some help! Not sure when even after a squash, it would not fix the problem!

PR #468, opened 2019-05-07
Generate new UA namespace nodes

kevinherron:
ObjectNode and VariableNodes now follow a different naming scheme.
Previously they assumed the type node they were modelling ended with a
"Type" suffix and tried to replace that with "Node". It turns out that
other information models do not follow this naming practice, so now the
code generator just appends "Node" to the BrowseName of the type node.


kevinherron:
Still need to generate new client nodes.

Issue #523, opened 2019-08-06
Client Redundancy via Subscription Transfer


kevinherron:
@kevinherron - would you be ok if I took a stab at implementing this as a PR and associating the commit with this enhancement request? 

I agree, the server would need to have diagnostics enabled, otherwise the client should return an error. 

Looking at the code, I think this enhancement can be achieved by adding logic to the _UaSubscriptionManager_/_OpcUaSubscriptionManager_ for transferring a subscription. This could:

1. Read the value of _ServerDiagnostics_/_EnabledFlag_ and return an error if false.
2. Call the _transferSubscriptions_ method in the client
3. Read the value of relevant _SubscriptionDiagnostics_ for the given subscription ids
4. Call _Server_/_GetMonitoredItems_ for each subscription id. This may be possible in parallel to the previous step.
5. Create an _OpcUaSubscription_ on the client for each subscription using the read diagnostics and add them to the _subscriptions_ Map in the subscription manager.
6. Using the results of the _GetMonitoredItems_ create the _OpcUaMonitoredItem_ for each subscription locally in the client. I still need to determine how I could provide all the fields for this object's definition, eg _ReadValueId_. One initial thought would be to set a _wasTransferred_ flag on each monitored item and if true when handling subscription notifications set the missing data from the information returned in the monitored item's notification. Other potential approaches could include _ResendData_ or _Republish_.
7. Setting the event/value consumer is a little tricky. I'll not be able to ascertain from the _GetMonitoredItems_ if the monitored item is a value or event. It may be necessary to set both for all transferred items, and then leverage the previously mentioned _wasTransferred_ flag to remove the extraneous consumer at runtime. 

Of course I'd be completely open to any suggestions, thoughts, collaborations, or other points of interest you could share. 

Is this something you'd be open to?

_Originally posted by @marcus-orchard in https://github.com/eclipse/milo/issues/189#issuecomment-518794750_

kevinherron:
I'm not convinced it has to be this complicated any more. Someone asked about this by email recently and I had a much shorter response:

> It does not seem possible to accomplish this right now, but it's close. The idea would be to use the TransferSubscription service to transfer the previous Subscriptions to the new Session established by the redundant client. All that's missing right now is some way to pass the UaSubscription objects from the old OpcUaSubscriptionManager object to the new one. Right now there is only immutable access to the subscription objects.

[...]

PR #536, opened 2019-08-19
Add support for client redundancy via subscription transfer (#523)

marcus-orchard:
Signed-off-by: Marcus Orchard <contactmarcusorchard@gmail.com>


kevinherron:
Hey Marcus,

It looks like you have an ECA on file for contactmarcusorchard@gmail.com but you've authored this commit with your Emerson email address.

marcus-orchard:
Oh crud. Ok, let me try to reauthor that.

marcus-orchard:
Alright, looks like everything is happy now.

kevinherron:
rebuild

kevinherron:
Thanks Marcus, this looks really good on the surface. I'm going to check it out and take a closer look sometime today. I might try to build a transfer example using it to get a feel for the API.

marcus-orchard:
Ok, just so you have my approach for reference (leaving out obvious null and empty checks):

1. Verify diagnostics are enabled by querying the `EnabledFlag` on the `ServerDiagnostics` node.
2. Call `getSessionDiagnosticsArray` on `SessionsDiagnosticsSummaryNode`.
3. Filter the results for items containing my expected session name, ordered by that name and select the head (name is appended with tick stamp).
4. Then I call `getObjectNode` on the client address space and provide the `sessionId` of the session I just retrieved and `classOf[SessionDiagnosticsObjectNode]`  as the arguments.
5. Finally, I call `getSubscriptionDiagnosticsArray` on the object node returned.

Now I have subscription diagnostic information for each subscription I want to transfer.
1. From here I can call `GetMonitoredItems` for each subscription.
2. I take the results of that and get the first client handle, then lookup the subscription configuration associated with that client handle. Now I know which server subscription id matches which of my subscription configurations.
3. Then I construct the `TransferMonitoredItemsRequest` for each grouping of monitoredItem/notification handlers that I need to reconstruct. 
4. This also requires me to create the `MonitoredItemTransferRequest`s for each item, using the earlier `GetMonitoredItems` results for server handles and and my configuration.
5. Finally I call `transferSubscription` on the `UaSubscriptionManager` passing in the information from my subscription diagnostics, subscription config, and the constructed `TransferMonitoredItemRequests`.

I think I wrote this in a way to allow other approaches as needed by other implementations, but having mine as reference I thought my be helpful.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscription.java
```diff
@@ -200,12 +203,66 @@ public OpcUaSubscription(
...
+                        if (req.getItemTransferCallback().isPresent()) {
+                            ItemTransferredCallback callback = req.getItemTransferCallback().get();
+                            for (UaMonitoredItem item : items) {
+                                int clientHandle = item.getClientHandle().intValue();
+                                callback.onItemTransferred(client.getSerializationContext(), item, clientHandle);
+                            }
+                        }
+                    });
+                } finally {
+                    permit.release();
```
kevinherron: Need to figure out a safer way to ensure `permit.release()` is called. If `modifyMonitoredItems()` call fails then the whole `thenApply` block never even runs.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscription.java
```diff
@@ -200,12 +203,66 @@ public OpcUaSubscription(
...
+                    monitoredItemsByRequest.get(transferReq).add(item);
+                    modifyRequests.add(new MonitoredItemModifyRequest(
+                        monitoredItem.getServerHandle(),
+                        monitoredItem.getRequestedParameters()));
+                }
+            }
+            CompletableFuture<List<StatusCode>> future =
+                this.modifyMonitoredItems(TimestampsToReturn.Neither, modifyRequests);
+
+            return future.thenApply(modifyResponse -> {
```
kevinherron: Not doing anything with the results of each item modification from `modifyResponse`. Should failure to modify the item be treated as a failure to "transfer" the item? Notify the callback and let them decide? Delete the item?

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscription.java
kevinherron: Something like this:
```java
CompletableFuture<List<UaMonitoredItem>> modifyFuture = new CompletableFuture<>();

modifyMonitoredItems(TimestampsToReturn.Neither, modifyRequests).thenApply(modifyResponse -> {
    monitoredItemsByRequest.forEach((req, items) -> {
        if (req.getItemTransferCallback().isPresent()) {
            ItemTransferredCallback callback = req.getItemTransferCallback().get();
            for (UaMonitoredItem item : items) {
                int clientHandle = item.getClientHandle().intValue();
                callback.onItemTransferred(client.getSerializationContext(), item, clientHandle);
            }
        }
    });

    return monitoredItemsByRequest.values()
        .stream()
        .flatMap(Collection::stream)
        .collect(toList());
});

return modifyFuture.whenComplete((r, ex) -> permit.release());
```

kevinherron: Something like this maybe:
```java
return modifyMonitoredItems(TimestampsToReturn.Neither, modifyRequests)
    .thenApply(modifyResponse -> {
        monitoredItemsByRequest.forEach((req, items) -> {
            if (req.getItemTransferCallback().isPresent()) {
                ItemTransferredCallback callback = req.getItemTransferCallback().get();
                for (UaMonitoredItem item : items) {
                    int clientHandle = item.getClientHandle().intValue();
                    callback.onItemTransferred(client.getSerializationContext(), item, clientHandle);
                }
            }
        });

        return monitoredItemsByRequest.values()
            .stream()
            .flatMap(Collection::stream)
            .collect(toList());
    })
    .whenComplete((i, ex) -> permit.release());
```

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscription.java
```diff
@@ -200,12 +203,66 @@ public OpcUaSubscription(
...
+                for (MonitoredItemTransferRequest monitoredItem : transferReq.getMonitoredItems()) {
+                    OpcUaMonitoredItem item = new OpcUaMonitoredItem(
+                        client,
+                        monitoredItem.getClientHandle(),
+                        monitoredItem.getItemToMonitor(),
+                        monitoredItem.getServerHandle(),
+                        new StatusCode(StatusCodes.Uncertain_InitialValue),
+                        monitoredItem.getRequestedParameters().getSamplingInterval(),
+                        monitoredItem.getRequestedParameters().getQueueSize(),
+                        monitoredItem.getRequestedParameters().getFilter(),
```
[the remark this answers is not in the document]
marcus-orchard: Still I think you are correct, I'll make that change.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscription.java
marcus-orchard: I think we should let each implementation decide, maybe? So, return a new object type of `TransferMonitoredItemResponse` comprised of a `List<UaMonitoredItems>` as well as the `List<StatusCode>` from the modify results?

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscription.java
marcus-orchard: This seems reasonable, I'll test it with the other changes mentioned in this review.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/TransferMonitoredItemsRequest.java
```diff
@@ -0,0 +1,53 @@
...
+    private final List<MonitoredItemTransferRequest> monitoredItems;
+    private final BiConsumer<UaMonitoredItem, Integer> itemTransferCallback;
+
+    public TransferMonitoredItemsRequest() {
+        this.monitoredItems = null;
+        this.itemTransferCallback = null;
+    }
+
+    public TransferMonitoredItemsRequest(
+        List<MonitoredItemTransferRequest> monitoredItems, BiConsumer<UaMonitoredItem,
```
marcus-orchard: Looks like autoformat did a number on here; I'm gonna put the BiConsumer on it's own newline.

kevinherron:
I want to see if I can piggy back on this to add a "local" transfer API as well. That way there's an easy to use transfer for when you're transferring on the same/local machine and have `UaSubscription` and `UaMonitoredItems` on hand, and this more sophisticated "remote" transfer option for transferring to a remote machine.

If there's not an easy to use "local" transfer API I'm going to have to spend a lot of time trying to support and explain how to use this "remote" transfer API.

marcus-orchard:
A separate "local" transfer makes sense to me, no need for some of the overhead that the "remote" transfer requires.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/MonitoredItemTransferRequest.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class MonitoredItemTransferRequest {
+    private final UInteger clientHandle;
+    private UInteger serverHandle;
+
+    private final ReadValueId itemToMonitor;
+    private final MonitoringMode monitoringMode;
+    private final MonitoringParameters requestedParameters;
+
+    public MonitoredItemTransferRequest() {
```
kevinherron: Do you need these zero-arg constructors? I'd like to delete them if not.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/TransferMonitoredItemResponse.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.eclipse.milo.opcua.sdk.client.api.subscriptions.UaMonitoredItem;
+import org.eclipse.milo.opcua.stack.core.types.builtin.StatusCode;
+
+public class TransferMonitoredItemResponse {
+
+    private final List<UaMonitoredItem> monitoredItems;
+    private final List<StatusCode> statusCodes;
+
+    public TransferMonitoredItemResponse() {
```
kevinherron: Do you need these zero-arg constructors? I'd like to delete them if not.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/MonitoredItemTransferRequest.java
```diff
@@ -0,0 +1,69 @@
...
+package org.eclipse.milo.opcua.sdk.client.subscriptions;
+
+import org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.UInteger;
+import org.eclipse.milo.opcua.stack.core.types.enumerated.MonitoringMode;
+import org.eclipse.milo.opcua.stack.core.types.structured.MonitoringParameters;
+import org.eclipse.milo.opcua.stack.core.types.structured.ReadValueId;
+
+public class MonitoredItemTransferRequest {
+    private final UInteger clientHandle;
+    private UInteger serverHandle;
```
kevinherron: `serverHandle` can be final too.

kevinherron:
@marcus-orchard do you mind if I start pushing some minor changes to this PR as I review?

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/MonitoredItemTransferRequest.java
marcus-orchard: Not at all. I add them out of habit from my c# days where some serializers needed them but would give cryptic errors when they were not there.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/TransferMonitoredItemResponse.java
marcus-orchard: Not at all. I add them out of habit from my c# days where some serializers needed them but would give cryptic errors when they were not there.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/MonitoredItemTransferRequest.java
marcus-orchard: Yep, missed that one.

marcus-orchard:
> @marcus-orchard do you mind if I start pushing some minor changes to this PR as I review?

Not at all. Do you want to address the items you mentioned in your review as well, or would you like me to tackle those? 

I think `TransferMonitorItemResponse` also needs to be refactored to `TransferMonitorItemsResponse`.

kevinherron:
I'll make the changes, I've got it open right now. 

I've intentionally not opted into Java's broken serialization system (and gone towards immutable objects with final fields), so no-arg constructors aren't necessary or desirable.

kevinherron:
I like the "trick" where you call modify when "transferring" monitored items to make sure you have current values for all the revised fields. I'm wondering if that same trick might be worth doing for subscriptions, even though you can technically get that information from the diagnostics. If an application configuration had all the other original/requested subscription values it might eliminate the need to read from the diagnostics?

marcus-orchard:
So, I did consider this at one point.

I think where I abandoned the idea is when I realized I have to query the subscription diagnostics anyways to get the subscription Ids to transfer, so I just grab the diagnostics when I'm querying that, and calling `modifySubscription` would be more overhead.

What initially interested me in this approach is that it reduces the information you need to provide when transferring a subscription, leading me to entertain the idea that it could potentially satisfy a larger number of unknown use cases.

But, another part of me thinks, "if they have a means to know the subscriptionId that the server assigned, I don't know why they wouldn't have the means to also know publishingInterval, lifetimeCount, etc." ; and if that is true then, again, calling `modifySubscription` would be just more overhead. 

Does that make sense? I did consider this as an option, but I'm open to the idea that I missed something that moves the needle more toward using the `modifySubscription` approach.

marcus-orchard:
Maybe to be safe we add an overload that uses this approach on the off chance there is a customer that somehow knows which `subscriptionId`'s they need to transfer without using diagnostics and  that only has access to the `subscriptionId` and their config?

kevinherron:
I get what you're saying. I forgot that you don't even have subscription ids as the client taking over.

An overload might not be a bad idea, I'll think about it a bit. 

If a client *did* somehow have subscription ids, or if there were some method that could be called to obtain subscription ids (not sure how that would work... call a method to get existing session with the same user, then call a method against the most recent session id to get existing subscriptions?), it might be possible to do a transfer without relying on diagnostics.

What I'm getting at is, while right now it appears diagnostics are required, I'm wondering if we can learn something from implementing this and I can use it to make some suggestions for OPC UA 1.05 on how to make client redundancy and subscription transfer suck less.

kevinherron:
@marcus-orchard what do you think about taking the `transferMonitoredItems` methods off of the `UaSubscription` interface and just giving them local visibility? Is there some scenario where a user might be calling these themselves rather than letting `UaSubscriptionManager` handle the transfer of the subscription and monitored items? Making these visible to the user seems like it might increase the odds of misuse.

kevinherron:
TODO: my local transfer implementation is still slightly broken because it doesn't offer the caller a chance to hook up monitored item event and value callbacks before the notification semaphore is released.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/api/subscriptions/UaSubscription.java
```diff
@@ -272,6 +301,12 @@
 
     }
 
+    interface ItemTransferredCallback {
+
+        void onItemTransferred(SerializationContext serializationContext, UaMonitoredItem item, int clientHandle);
```
kevinherron: Is the `clientHandle` parameter necessary? Can't you get it just by calling `item.getClientHandle()`?

marcus-orchard: Yeah, you're right. 

I modeled this off of the `createMonitoredItems` method, but didn't have a use case to provide an index like that method does. Do you think we need an index there? If not I'm good with just removing the `clientHandle`.

marcus-orchard:
> @marcus-orchard what do you think about taking the `transferMonitoredItems` methods off of the `UaSubscription` interface and just giving them local visibility in OpcUaSubscription?
> 
> Is there some scenario where a user might be calling these themselves rather than letting `UaSubscriptionManager` handle the transfer of the subscription and monitored items? Making these visible to the user seems like it might increase the odds of misuse.

Agreed.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/api/subscriptions/UaSubscription.java
kevinherron: Modified callback so it provides an index instead of client handle, but still not sure it's actually useful for anything.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscriptionManager.java
```diff
@@ -351,6 +351,71 @@ private UInteger getLifetimeCount(double publishingInterval, UInteger maxKeepAli
...
+
+        subscriptions.put(newSubscription.getSubscriptionId(), newSubscription);
+
+        return newSubscription
+            .transferMonitoredItems(subscription, callback)
+            .thenApply(v -> newSubscription);
+    }
+
+    @Override
+    public CompletableFuture<UaSubscription> transferSubscription(
```
kevinherron: I think this needs a `nextSequenceNumber` parameter, obtained from a `SubscriptionDiganosticsDataType` for the corresponding subscription. Without it the subscription manager will try to call republish for potentially many thousands of sequence numbers after receiving the first `PublishResponse` for the transferred subscription.

marcus-orchard: I think the `transferSubscriptions` method returns `availableSequenceNumbers`. Since this is always called no matter the approach I think this would be better (assuming I'm on the right track here). Would the first first item in that array be sufficient? 

We would set the `lastSequenceNumber` value of the subscription with that number minus 1?

kevinherron:
Hey, just wanted to let you know I haven't forgotten this, I've just been wildly busy the last couple weeks at both work and home. Things should calm down after next week.

marcus-orchard:
No worries at all and I was worried you thought the same about me. I'll keep a look out for activity and keep pressing on with other things.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscriptionManager.java
```diff
@@ -351,6 +351,79 @@ private UInteger getLifetimeCount(double publishingInterval, UInteger maxKeepAli
...
+        double requestedPublishingInterval,
+        double publishingInterval,
+        UInteger requestedLifetimeCount,
+        UInteger lifetimeCount,
+        UInteger requestedMaxKeepAliveCount,
+        UInteger maxKeepAliveCount,
+        UInteger maxNotificationsPerPublish,
+        boolean publishingEnabled,
+        UByte priority,
+        List<TransferMonitoredItemsRequest> itemsToTransfer
```
kevinherron: I'm not convinced `itemsToTransfer` should be a List of `TransferMonitoredItemsRequest`. It already contains what should be all the monitored items for the subscription being transferred inside of it (`List<MonitoredItemTransferRequest>`).

Can you convince me otherwise @marcus-orchard ?

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscription.java
```diff
@@ -301,6 +306,123 @@ public OpcUaSubscription(
...
+     * <p>
+     * This subscription shall already have been transferred to this client's session via the Transfer Subscription
+     * service.
+     *
+     * @param transferRequests a list of {@link TransferMonitoredItemsRequest}s.
+     * @return a {@link TransferMonitoredItemsResponse} that includes a list of the {@link UaMonitoredItem}s
+     * and a list of the {@link StatusCode}s for each modify result.
+     */
+    CompletableFuture<TransferMonitoredItemsResponse> transferMonitoredItems(
+        List<TransferMonitoredItemsRequest> transferRequests
```
kevinherron: `transferRequests` could/should just be a single `TransferMonitoredItemsRequest`, depending on what we do on `OpcUaSubscriptionManager#transferSubscription`. See the review comment for more info.

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscriptionManager.java
marcus-orchard: @kevinherron - Sorry, was tied up on another project. Started looking at your comment this morning. Should be able to follow up shortly.

marcus-orchard: OK, took me a few minutes but I remember my thinking in regards to this. 

I wanted to allow the ability to *easily* support different callbacks for groups of monitored items within a subscription should the original subscription have been created that way (or for some other unforeseen need).

``` 
    private final List<MonitoredItemTransferRequest> monitoredItems;
    private final ItemTransferredCallback itemTransferCallback;
```

opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/subscriptions/OpcUaSubscription.java
marcus-orchard: I believe this is the same situation as the above conversation where the intention is to allow different `ItemTransferredCallback`s for different groups of `MonitoredItemTransferRequest`s

kevinherron: Is this a scenario you actually need or is it theoretical? It seems to add considerable overhead to the API signature and complexity.

kevinherron: Also consider that maybe this scenario could be accomplished with multiple transfer calls instead.

marcus-orchard: Looking at my code it appears that initially I did, but then refactored the code to utilize a single callback that incorporates a pattern match to handle the different scenarios.

kevinherron:
@FSangouard I don't know... it will take some work to make it mergeable and then try to understand it and make sure it's (still) sane. 

I probably won't be able to take a good look at this until after I attend the OPC interoperability event in October. In the mean time I am working hard towards OPC UA 1.05 support that I plan to test at the event.

Issue #584, opened 2020-01-06
Automatic-Module-Name needs to be unique for each maven module


comtel2000:
based on #442 please choose a dedicated automatic module name for each maven module (stack-client, stack-core, sdk-client, etc.).

current 'stack-core' MANIFEST.MF:

```
..
Automatic-Module-Name: org.eclipse.milo.opcua
Bundle-License: https://www.eclipse.org/org/documents/epl-2.0/EPL-2.0.
 html
Bundle-SymbolicName: org.eclipse.milo.stack-core
Implementation-Version: 0.3.6
..
```
needs to be changed into:
```
..
Automatic-Module-Name: org.eclipse.milo.stack.core
..
```

Just like the Bundle-SymbolicName just with '.' instead of '-'

comtel2000:
or changed to the root package name (stack-core):
```
..
Automatic-Module-Name: org.eclipse.milo.opcua.stack.core
..
```

kevinherron:
I think I'll just use the root package name. Should be able to do this and release it with 0.3.7 this week.

comtel2000:
Thanks, that's really quick :)

PR #585, opened 2020-01-07
Add automatic module names (#584)

comtel2000:
works for me, BR, Joerg

### DIFF
--- a/milo-examples/client-examples/pom.xml
+++ b/milo-examples/client-examples/pom.xml
@@ -10,6 +10,10 @@
 
     <artifactId>client-examples</artifactId>
 
+    <properties>
+        <javaModuleName>org.eclipse.milo.examples.client</javaModuleName>
+    </properties>
+
     <profiles>
         <profile>
             <id>standalone-client</id>
--- a/opc-ua-sdk/sdk-client/pom.xml
+++ b/opc-ua-sdk/sdk-client/pom.xml
@@ -10,6 +10,10 @@
 
     <artifactId>sdk-client</artifactId>
 
+    <properties>
+        <javaModuleName>org.eclipse.milo.opcua.sdk.client</javaModuleName>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.eclipse.milo</groupId>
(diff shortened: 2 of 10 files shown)


kevinherron:
I think `bsd-parser` will need one as well?

comtel2000:
you're right. Each 'jar' module requires a unique automatic name.
I missed **bsd-parser-gson** and **bsd-parser-core**

The **bsd-parser-core** should get a sub-package 'core' to be safe for JPMS.
I can rename/ or move it, but I don't know how many dependencies must be update too..
What do you think?

kevinherron:
These packages have already started being reorganized on this branch in 0.4: https://github.com/eclipse/milo/tree/0.4-dtd-updates

comtel2000:
with the 0.4-dtd-updates branch it will be more complex. the 'core' only contains the resource xsd file and core classes are moved to the previous main module?
The old structure (with a additional sub-package) looks better to get it run with the java module system.

kevinherron:
In that branch all the code in `bsd-core` is generated at build time under the package `org.opcfoundation.opcua.binaryschema`.

comtel2000:
ok, I need some time to check that. Once a module name is set, it's a pain in the a... to modify it later on.. :)

kevinherron:
Ok. Any suggestions for how we can get this working for 0.3.7 without changing package names? Or should we wait and do the automatic module names in 0.4 instead?

I'm a little disoriented having just got back to work from a ~3 week holiday, but if I remember correctly I have a few branches off of dev/0.4 going and they're all pretty close to done, so a 0.4 release isn't too far in the future. Need to dig in this week and find out where everything stands.

comtel2000:
I guess this will work now for 3.7. For the 4.0 branches and JPMS compatibility the module root packages should be unique but as long as it's not switched to fully support JPMS (module-classpath) it works also with the current state (automatic-module-names).

kevinherron:
Okay, thanks for doing this 👍

kevinherron:
If this is ready and looks good to you I'll merge it.

comtel2000:
perfect, thx

PR #638, opened 2020-05-03
New ManagedSubscription API

kevinherron:


sarthak-k:
How is this different from normal subscription API? Any benefits of using this new API?

kevinherron:
@sarthak-k it's an easier to use API that caters to the most common uses and has both blocking and non-blocking methods, with a focus on blocking.

Please use the mailing list instead of reviving an old issue for questions like this.

Issue #696, opened 2020-09-16
MinimumSamplingInterval attribute in generated client variable nodes can't be null


kevinherron:
The new generated client variable nodes mistakenly use primitive `double` instead of `Double` for the MinimumSamplingInterval attribute, making it impossible to use `null` when this optional attribute doesn't exist.

PR #697, opened 2020-09-16
Handle null optional attributes and unexpected types

### LINKED ISSUE #694: OPC-UA client fails for getting variable node, connecting to kepware server
**Describe the bug**
I'm trying to connect to kepware server simulation node using the OPC UA client 0.5.0 SDK
the connection succeeds and I'm able to write value to it.
when I try to use this connection to get the variable node that I wrote to it give me an error saying that
`UaException: status=Bad_UnexpectedError, message=java.util.concurrent.ExecutionException: java.lang.ClassCastException: org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.UInteger cannot be cast to java.lang.Double`
up on inspecting it further it is seen that the sampling interval is by default set to 10 and it fails when ittried to get the variable node attributes for 
`        Double minimumSamplingInterval = (Double) attributeValues.get(13).getValue().getValue();`
as this is 10 and it gets a value of 10 as Uint.

**Expected behavior**
Since I was able to get the node and read the value using the previous SDK I should be able to get the node without any exception as nothing has changed

**Logs and Packet Captures**
```
Caused by: java.lang.ClassCastException: org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.UInteger cannot be cast to java.lang.Double
	at org.eclipse.milo.opcua.sdk.client.AddressSpace.newVariableNode(AddressSpace.java:1221)
	at org.eclipse.milo.opcua.sdk.client.AddressSpace.lambda$getVariableNodeAsync$4(AddressSpace.java:310)
	at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:602)
	at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:577)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
	at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
	at org.eclipse.milo.opcua.stack.client.UaStackClient.lambda$deliverResponse$5(UaStackClient.java:256)
	at org.eclipse.milo.opcua.stack.core.util.ExecutionQueue$Task.run(ExecutionQueue.java:119)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

**Additional context**
below is the capture from ua client tool
image

kevinherron:
fixes #694 and #696

### DIFF
--- a/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/AddressSpace.java
+++ b/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/AddressSpace.java
@@ -20,6 +20,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
@@ -43,6 +44,7 @@ import org.eclipse.milo.opcua.stack.core.types.builtin.ExpandedNodeId;
 import org.eclipse.milo.opcua.stack.core.types.builtin.LocalizedText;
 import org.eclipse.milo.opcua.stack.core.types.builtin.NodeId;
 import org.eclipse.milo.opcua.stack.core.types.builtin.QualifiedName;
+import org.eclipse.milo.opcua.stack.core.types.builtin.StatusCode;
 import org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.UByte;
 import org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.UInteger;
 import org.eclipse.milo.opcua.stack.core.types.enumerated.BrowseDirection;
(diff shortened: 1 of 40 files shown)


opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/AddressSpace.java
```diff
@@ -1313,6 +1315,23 @@ private UaViewNode newViewNode(NodeId nodeId, List<DataValue> attributeValues) {
...
+
+        if (statusCode != null && statusCode.isBad()) {
+            return null;
+        } else {
+            Object attributeValue = value.getValue().getValue();
+
+            try {
+                return attributeClazz.cast(attributeValue);
+            } catch (ClassCastException e) {
+                return null;
```
sushantkale90, line 1330: since the value may exist here so, can we try and return the value instead of null.

in my case it was Uint(10) instead of Double so can we cast it to `Double.valueOf(i)` else return null.

@kevinherron

kevinherron: I'm not interested in adding logic to every attribute read for coercing values of incorrect types from a defective server into the correct type. It's enough that we don't throw an uncaught exception here any more.

PR #716, opened 2020-10-22
Allow extending OpcUaServer functionality via subclassing

balazsgrill:
Signed-off-by: Balazs Bea <balazs.bea@incquerylabs.com>

- [ ] You have signed an Eclipse Contributor Agreement and are committing using the same email address
- [ ] Your code contains any tests relevant to the problem you are solving
- [ ] All new and existing tests passed
- [ ] Code follows the style guidelines and Checkstyle passes

See [CONTRIBUTING](https://github.com/eclipse/milo/blob/master/CONTRIBUTING.md) for more information.

### DIFF
--- a/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/OpcUaServer.java
+++ b/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/OpcUaServer.java
@@ -75,17 +75,17 @@ public class OpcUaServer {
 
     private final Map<UInteger, Subscription> subscriptions = Maps.newConcurrentMap();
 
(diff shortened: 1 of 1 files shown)


kevinherron:
I think we should talk more by email. I see very little benefit in these changes because none of objects being instantiated are extensible - they are all concrete implementations that you cannot provide a sensible alternative for via these factory methods.

I do have plans to make the server and services as well as session and subscription management extensible somehow but in an eventual version 2.0 of the library.

There is only perhaps one more release (0.6.x) before I'd like to move to a stable 1.0 release. Now is not the time for major changes or features like making all of these things extensible. 

Currently the focus is on:
- fixing any remaining bugs
- ensuring that there is no missing functionality that would require API changes
- documentation

PR #795, opened 2021-03-17
Update netty-channel-fsm to version 0.5

kevinherron:
This version and its transitive dependency strict-machine 0.4 both include
Automatic-Module-Name entries in their MANIFEST.MF files.


kevinherron:
rebuild

kevinherron:
rebuild

kevinherron:
rebuild

PR #949, opened 2022-03-24
allow all-zero nonces when no nonce is required

lovasoa:
See https://github.com/locka99/opcua/issues/182

Rust's [opcua-server](https://crates.io/crates/opcua-server) v0.9.0 sends an all-zero nonce for security policies where no nonce is required.
There is no reason to throw an error in this case, since the nonce is not used anyway.

### DIFF
--- a/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/util/NonceUtil.java
+++ b/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/util/NonceUtil.java
@@ -232,7 +232,7 @@ public class NonceUtil {
             );
         }
 
-        if (bs.length > 0 && Arrays.areAllZeroes(bs, 0, bs.length)) {
+        if (minimumLength > 0 && bs.length > 0 && Arrays.areAllZeroes(bs, 0, bs.length)) {
             throw new UaException(StatusCodes.Bad_NonceInvalid, "nonce must be non-zero");
         }
     }


kevinherron:
I need to dig through the mailing lists, but this case was explicitly called out as something stacks should not allow when discussing CVE-2019-19135.

Even when then transport security is None, if a username and password is specified, and both client and server have certificates, then the password is encrypted before it is sent. The nonce is appended to the password before it is encrypted. A fixed nonce of all zeroes is just as bad as a 0-length nonce in terms of vulnerability to a replay attack.

You should probably not be using an all-zero nonce in any cirumstance.

kevinherron:
Further, if you are seeing nonce validation fail in any version of Milo from 0.3.7 onward, then it is probably a case where the nonce is important. The validation was relaxed to only situations where the nonce matters in https://github.com/eclipse/milo/commit/35ee06ea87676db9fdf4caf748b9265c40d1c841

lovasoa:
Ok, thank you for that information ! I'll update my pull request to rust's [opcua-server](https://crates.io/crates/opcua-server)

kevinherron:
I'm digging around in the spec a little bit here - after looking at the Wireshark capture in the other issue I see this validation failure is occurring rather early, after receiving the OpenSecureChannelResponse. While this nonce _should_ be `null` when `SecurityPolicy.None` is being used, I think the client should also be ignoring it. 

It's a later nonce that validation is important for, but only when it's used to protect authentication on an otherwise unsecured channel.

lovasoa:
Indeed ! But you were mentioning recent changes to the library; I cannot guarantee that ignition uses the latest version of milo...

kevinherron:
> I cannot guarantee that ignition uses the latest version of milo...

I can :) I work at Inductive Automation and wrote the OPC UA code.

lovasoa:
Oh, great ! Then let's fix the bug in ignition :)

kevinherron:
I've opened https://github.com/eclipse/milo/issues/950 to further relax this. I'm going to fix it differently than your PR here, but thank you.

I would suggest you open a PR against the rust stack to use a null nonce as the spec suggests, instead of an all-zero nonce.

lovasoa:
Thank you for your reactivity !

I opened a pr to the rust implementation to always generate a 32-byte random nonce when SecurityPolicy is None, so that it can be used when needed for password encryption as you mentioned.

lovasoa:
There is something I don't understand. The CVE mentioned earlier said

> [...] all 3 of the following conditions are met:
>  *   SecurityPolicy is None
>  *  using username/password or X509-based authentication
>  *  the server **has a defect** causing it to send null/empty or zeroed nonces
 
But your other issue mentions the spec: 


> > The OpenSecureChannel Messages are not signed or encrypted if the SecurityMode is None. The Nonces are ignored and should be set to null. The SecureChannelId and the TokenId are still assigned but no security is applied to Messages exchanged via the channel. The SecurityToken shall still be renewed before the RevisedLifetime expires. Receivers shall still ignore invalid or expired TokenIds.
> 
> from https://reference.opcfoundation.org/v104/Core/docs/Part6/6.7.4/

So is it **a defect** or a standard-compliant behavior to send a null nonce ?

kevinherron:
> So is it a defect or a standard-compliant behavior to send a null nonce ?

It is only compliant when the null nonce is in the OpenSecureChannelResponse and the channels are not using security.

There are other nonce exchanges, in the CreateSessionResponse and ActivateSessionResponse, and in these it is important in regards to the CVE that the nonce be non-empty and non-zero if they are going to be used to subsequently encrypt credentials for transit on an otherwise unsecure channel.

kevinherron:
Let me know if you need anything else and thanks for the report and PR.

PR #1162, opened 2023-09-15
Add AttributeFilter `::readAttribute` and `::writeAttribute`

kevinherron:
These are now called by AttributeReader and AttributeWriter when responding to Read and Write service calls, and allow for filter implementations that can indicate a failure to get or set an attribute by throwing UaException, as opposed to the `getAttribute` and `setAttribute` methods that must not throw.

Additionally, filter contexts are simplified, removing the separate GetAttributeContext and SetAttributeContext in favor of just AttributeFilterContext.

### DIFF
--- a/milo-examples/server-examples/src/main/java/org/eclipse/milo/examples/server/AttributeLoggingFilter.java
+++ b/milo-examples/server-examples/src/main/java/org/eclipse/milo/examples/server/AttributeLoggingFilter.java
@@ -13,8 +13,7 @@ package org.eclipse.milo.examples.server;
 import java.util.function.Predicate;
 
 import org.eclipse.milo.opcua.sdk.server.nodes.filters.AttributeFilter;
-import org.eclipse.milo.opcua.sdk.server.nodes.filters.AttributeFilterContext.GetAttributeContext;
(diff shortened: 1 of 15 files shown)


kevinherron:
An example of how this might be useful:
```kotlin
/**
 * [AttributeFilter] that checks to see if a Node is an [AnalogItemType] and, if it is, throws an
 * exception with [StatusCodes.Bad_OutOfRange] when a value is written outside the configured EU
 * [org.eclipse.milo.opcua.stack.core.types.structured.Range], or [StatusCodes.Bad_TypeMismatch]
 * if the value is not a number.
 */
object EuRangeCheckFilter : AttributeFilter {

    override fun writeAttribute(
        ctx: AttributeFilterContext,
        attributeId: AttributeId,
        value: Any?
    ) {

        val node: UaNode = ctx.node

        if (attributeId == AttributeId.Value && node is AnalogItemType) {
            val v: Any? = (value as? DataValue)?.value?.value

            if (v is Number) {
                val low: Double = node.euRange.low
                val high: Double = node.euRange.high

                if (v.toDouble() !in low..high) {
                    throw UaException(StatusCodes.Bad_OutOfRange)
                }
            } else {
                throw UaException(StatusCodes.Bad_TypeMismatch)
            }
        }

        ctx.writeAttribute(attributeId, value)
    }

}
```

PR #1201, opened 2024-01-12
Updated Guava version to 32.1.3-jre

LocutusV0nB0rg:
This PR contains only the update of the guava version. It should make the warning about insecure transitive dependencies disappear, even though milo is basically unaffected.

- [ ok ] You have signed an Eclipse Contributor Agreement and are committing using the same email address
- [ ok ] Your code contains any tests relevant to the problem you are solving
- [ ok ] All new and existing tests passed
- [ ok ] Code follows the style guidelines and Checkstyle passes

### DIFF
--- a/opc-ua-stack/pom.xml
+++ b/opc-ua-stack/pom.xml
@@ -23,7 +23,7 @@
 
     <properties>
         <bouncycastle.version>1.75</bouncycastle.version>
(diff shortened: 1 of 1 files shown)


LocutusV0nB0rg:
Commited with bad email, will open another PR

PR #1255, opened 2024-03-31
Allow writing `null` when "AllowNulls" Property is `true`

### LINKED ISSUE #1254: Setting a node's value to null resp. clearing the node
**Describe the bug**
I am trying to clear a node's value by setting it to `null`. For this I used a `DataValue(null)` as the code suggests (see below):

```
client.writeValue(nodeId, DataValue.valueOnly(null)).get();
```

This leads to an Exception to be thrown as follows:

```
UaException: status=Bad_TypeMismatch, message=The value supplied for the attribute is not of the same type as the attribute's value.
        at org.eclipse.milo.opcua.sdk.server.util.AttributeWriter.validateDataType(AttributeWriter.java:215)
```

**Expected behavior**
I expected the node to be cleared instead.

**Additional context**
I am using milo 0.6.12.

I tried to trace down the error. What I found is:
* When sending a null value, the value is decoded into `DataValue(Variant(null))` here: https://github.com/eclipse/milo/blob/7f49b8d4cc4c1efb4a9141d2b87d1286eb3ba9e7/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/serialization/OpcUaBinaryStreamDecoder.java#L221
* The server checks for null here and throws an exception: https://github.com/eclipse/milo/blob/7f49b8d4cc4c1efb4a9141d2b87d1286eb3ba9e7/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/util/AttributeWriter.java#L215
* I think either the value should be set to `null` instead of `Variant(null)` in the first link, or it should just return `null` instead of throwing an exception in the second link.

Does that make sense?

kevinherron:
fixes #1254


markuzzi:
Looks good, this will solve my issue, thanks!

PR #1263, opened 2024-04-19
Migrate from TestNG to JUnit5

apupier:
- all compilation issues fixed
- remains few failing tests

- [x] You have signed an Eclipse Contributor Agreement and are committing using the same email address
- [ ] Code follows the style guidelines and Checkstyle passes

See [CONTRIBUTING](https://github.com/eclipse/milo/blob/master/CONTRIBUTING.md) for more information.

### DIFF
--- a/opc-ua-sdk/dictionary-reader/pom.xml
+++ b/opc-ua-sdk/dictionary-reader/pom.xml
@@ -34,8 +34,7 @@
 
         <dependency>
             <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-engine</artifactId>
-            <version>5.5.2</version>
+            <artifactId>junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/opc-ua-sdk/dictionary-reader/src/test/java/org/eclipse/milo/opcua/sdk/client/dtd/DataTypeDictionaryReaderTest.java
+++ b/opc-ua-sdk/dictionary-reader/src/test/java/org/eclipse/milo/opcua/sdk/client/dtd/DataTypeDictionaryReaderTest.java
@@ -124,7 +124,7 @@ class DataTypeDictionaryReaderTest {
             fragmentSize
         ).get();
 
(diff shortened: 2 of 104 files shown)


apupier:
Please note that I'm on PTO next week. Feel free to update/modify if needed

kevinherron:
Thanks for picking this one up. I'll try to take a closer look at this today or over the weekend.

On first glance it looks mostly good, except I think you missed the most annoying thing about the migration, which is that JUnit and TestNG assertions swap the position of the "expected" and "actual" parameters :crying_cat_face:

apupier:
Provided as a separate commit to more easily revert in case there is a problem for assertEquals using:

image

kevinherron:
Careful, I think you might be hitting existing JUnit tests and swapping the positions of _those_ assertions.

apupier:
> Careful, I think you might be hitting existing JUnit tests and swapping the positions of _those_ assertions.

hum, I didn't noticed that there were existing Junit tests. Effectively I hit them.
I need to stop for today and then I have a week of PTO.
Feel free to pick the commits you are interested in, or to start from scratch.

apupier:
and thanks for the quick feedback!

kevinherron:
Enjoy your PTO, thanks for the contribution.

kevinherron:
IntelliJ allows you to specify the scope of a Find (and Replace) action, and it looks like local changes is one of those scopes. I might be able to get something working using that... I'll play with it a bit.

kevinherron:
@protogenes that's another good solution, though it would have to be applied as the first step, before all the package migration from TestNG to JUnit happens.

kevinherron:
@VSSW-GTN thanks but I'm not interested in adding another library. I've never been convinced that assertion "DSLs" like this end up being worth it in the end. Regular assertions aren't difficult to read and don't come with the overhead of needing to understand the assertion language of whatever library you're using. Maybe if the tests were complex enough it would start to make sense.

kevinherron:
If somebody picks this up please submit the new PR against `dev/1.0`.

PR #1292, opened 2024-08-01
Return Bad_AttributeIdInvalid when reading InverseName on a symmetric…

### LINKED ISSUE #1291: Reading InverseName of a symmetric reference node should return Bad_AttributeIdInvalid
### Discussed in https://github.com/eclipse/milo/discussions/1288

<div type='discussions-op-text'>

<sup>Originally posted by **johannwesely** July 30, 2024</sup>
Hi,
I think i solved some problems regarding ctt and want to ask if my solutions seems legit and if you could add them to dev1.0?
I had an Issue with CTT Script Base Information->Base Info Type System-> ReferenceTypesValidation 
CTT ReferenceTypes 2024-07-30 145550

I added in AttributeReader Line 156 a symmetric check as mentioned in part 3,5.2

 
```
switch (attributeId) {
                case DataTypeDefinition:
                case RolePermissions:
                case UserRolePermissions:
                case AccessRestrictions:
                case AccessLevelEx:
                    if (value == null) {
                        return new DataValue(StatusCodes.Bad_AttributeIdInvalid);
                    }

            }
            if (attributeId == AttributeId.InverseName && node instanceof UaReferenceTypeNode referenceNode){
                if( referenceNode.getSymmetric()){
                    return new DataValue(StatusCodes.Bad_AttributeIdInvalid);
                }
            }
```

refrencetype green 2024-07-30 150114
and im green

Also I had Problem with ServerDiagnostics 
Diagnostics 2024-07-30 150423
here i set in OPcUaNamespace the ServerDiagnostic Enabled Flag in line 186 ` serverTypeNode.getServerDiagnosticsNode().setEnabledFlag(true);` 
now i have an new error(ConditionType not found) but Timestamps are correct now.

How is this flag meant to be set must the client first enable and then use diagnostics ? But Ctt doesn't do this in script.

Im Trying to figure out more CTT failure solutions, but i dont know if the Forum is the right place to post them?

Kind Regards 
Johann Wesely</div>

kevinherron:
… reference

fixes #1291

### DIFF
--- a/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/AttributeReader.java
+++ b/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/AttributeReader.java
@@ -135,10 +135,10 @@ public class AttributeReader {
                 return dvb.applyTimestamps(attributeId, timestamps).build();
             }
         } else {
-            // These attributes are either structures or primitive types, and should
-            // not expose a null value to clients, so if they are null in the Node
-            // that means they are not implemented/supported for the node, and we need
-            // to return Bad_AttributeIdInvalid.
+            // These attributes are either structures/primitive types, and should not expose a
+            // null value to clients, or they are optional attributes.
+            // In either case a null value means that they are not implemented/supported for the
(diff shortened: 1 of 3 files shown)


kevinherron:
rebuild

PR #1317, opened 2024-09-07
Refactor to a common `CertificateValidator` interface

kevinherron:
A lot of refactoring to support a common `CertificateValidator` interface, and to accept `CertificateValidator` in the constructor of `DefaultApplicationGroup`. This enables application groups to be constructed that aren't forced into a validator implementation, allowing e.g. client, server, insecure, or other custom validators to be used.


kevinherron:
rebuild

kevinherron:
rebuild

PR #1348, opened 2024-11-28
Notify SessionActivityListeners using client executor

### LINKED ISSUE #1300: Execution of SessionFsm transition actions can create a deadlock
**Describe the bug**
SessionFsm transition actions are not properly executed in order, and it can cause a deadlock.

The main use case is resubscribing subscriptions, that were failed to transfer to a new session.
There are 2 main points on the client side, when we are notified to do it:

1. SessionActivityListener.onSessionActive - it is assumed that we are safe to work with anything session related, lik subscribing to nodes
2. ManagedSubscription.StatusListener.onSubscriptionTransferFailed - it is assumed here, that everything can be done with the SubscriptionManager here.

However, this is not the case, we can be in a deadlock easily. 

The main problem is, that onSessionActive and onSubscriptionTransferFailed callbacks can run before the session's SessionFuture (with the key KEY_SESSION_FUTURE) is properly resolved.

The reason is, that the SessionFsm is set up in SessionFsmFactory so that the session's future is only resolved when the InitializeSuccess event is called between the Intializing and Active state transition, see https://github.com/eclipse/milo/blob/master/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/session/SessionFsmFactory.java#L513

It resolves the session's future via the executorservice, and not directly, see line 565.

However, the onSessionActive callbacks are called directly, when the state changes to Active from any other state, in https://github.com/eclipse/milo/blob/master/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/session/SessionFsmFactory.java#L587

Since these callse are not made through the client's executorservice, they can happen-before the sessionFuture is resolved.

So there is no guarantee, that even after onSessionActive is called, anything can be done, high-level APIs like ManagedSubscription.create() will hang indefinitely, as they are in a deadlock : they are waiting for the future's resolution, but that is scheduled later.

I think, the issue is the same as https://github.com/eclipse/milo/issues/614

In that issue, you mentioned that: 
> I'll start looking at whether I can make these callbacks safer. In the meantime, when you're executing inside a callback, whether from the subscription manager or just a CompletableFuture for a response, just make sure not to block waiting for another response on the same thread.

What is the preferred way to solve this issue? When and how can ManagedSubscription.create() (and any reentrant calls to getSession()) safely called?

Also, it needs to be documented properly in the Javadocs, and perhaps in the milo-examples code.

**Expected behavior**
SessionFsm transition actions have a well-defined order.

**Additional context**
Milo version 0.6.14, but affects all others.

kevinherron:
This avoids deadlock when the user makes a blocking call that requires the Session from the SessionFsm from inside the callback.

fixes #1300

### DIFF
--- a/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/session/SessionFsmFactory.java
+++ b/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/session/SessionFsmFactory.java
@@ -593,8 +593,11 @@ public class SessionFsmFactory {
                 SessionFsm.SessionActivityListeners sessionActivityListeners =
                     KEY_SESSION_ACTIVITY_LISTENERS.get(ctx);
 
-                sessionActivityListeners.sessionActivityListeners
-                    .forEach(listener -> listener.onSessionActive(session));
+                client.getConfig().getExecutor().execute(
+                    () ->
+                        sessionActivityListeners.sessionActivityListeners
+                            .forEach(listener -> listener.onSessionActive(session))
+                );
             });
 
         // onSessionInactive() callbacks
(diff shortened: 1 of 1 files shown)


kevinherron:
rebuild

kevinherron:
rebuild

kevinherron:
rebuild

Issue #1390, opened 2025-03-26
Encoded DataValue should not explicitly include Good StatusCode


kevinherron:
When a `DataValue` with `null` or `StatusCode.GOOD` is encoded, the EncodingMask field should not set the bit indicating StatusCode is included, and the StatusCode value should not be encoded inline.

---
As it's implemented now, when `null` the bit _is_ set, and an explicit StatusCode value of 0 is encoded: https://github.com/eclipse-milo/milo/blob/c6a875c2fa8f65c48815410d6bbf02d43560bcd4/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/serialization/OpcUaBinaryStreamEncoder.java#L291-L293

<img width="545" alt="Image" src="https://github.com/user-attachments/assets/fb12468f-f9bc-46d0-a65e-88dec8f3ab06" />

PR #1393, opened 2025-03-26
Don't explicitly encode `null` StatusCode in DataValue

kevinherron:
A DataValue constructed with a `null` StatusCode is semantically equivalent to one with a Good (0) StatusCode, and shouldn't result in the corresponding bit in the DataValue EncodingMask field being set and an explicit 0 value being encoded.

forward port of #1390


kevinherron:
Going to disallow `null` StatusCode in DataValue for 1.0 instead.

PR #1394, opened 2025-03-26
Disallow `null` for StatusCode field in DataValue

kevinherron:
A `null` value is semantically the same, and encodes the same, as `StatusCodes.GOOD`. Any distinction is lost during serialization.


kevinherron:
This fixes the issue described by #1390 for the `dev/1.0` branch.

PR #1401, opened 2025-04-02
Set `lastSequenceNumber` on the first PublishResponse

kevinherron:
Set `lastSequenceNumber` even if the first PublishResponse is empty (keep-alive). 

This prevents an unnecessary Republish call when a Server sends an empty (keep-alive) first PublishResponse followed by a second PublishResponse once it has acquired initial values for the MonitoredItems in the Subscription. This is explicitly allowed by the specification.


kevinherron:
Passage from the spec:

> When a Subscription is created, the first Message is sent at the end of the first publishing cycle to inform the Client that the Subscription is operational. A NotificationMessage is sent if there are Notifications ready to be reported. If there are none, a keep-alive Message is sent instead that contains a sequence number of 1, indicating that the first NotificationMessage has not yet been sent. This is the only time a keep-alive Message is sent without waiting for the maximum keep-alive count to be reached, as specified in (f) above.

PR #1406, opened 2025-04-07
Update UaDataType and DataTypeCodec interfaces

kevinherron:
Adds `DataTypeCodec::getNamespaceUri`, `DataTypeCodec::getEncodingName`, and `UaDataType::getEncodingName`.

These provide encoders with the necessary information to safely encode type names and to derive namespace URIs from either a DataType NodeId or a DataTypeCodec.

The rest of the (many) changes are just fallout from adding/implementing these new methods.

### DIFF
--- a/milo-examples/server-examples/src/main/java/org/eclipse/milo/examples/server/ExampleNamespace.java
+++ b/milo-examples/server-examples/src/main/java/org/eclipse/milo/examples/server/ExampleNamespace.java
@@ -1023,7 +1023,11 @@ public class ExampleNamespace extends ManagedNamespaceWithLifecycle {
         .getServer()
         .getDynamicDataTypeManager()
         .registerType(
-            dataTypeId, DynamicCodecFactory.create(dataType, dataTypeTree), null, null, null);
+            dataTypeId,
+            DynamicCodecFactory.create(NAMESPACE_URI, dataType, dataTypeTree),
+            null,
+            null,
+            null);
 
     return dataType;
   }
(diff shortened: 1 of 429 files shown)


kevinherron:
@johannwesely this will probably require changes for you since you have custom types, but they're minor. The diff for this PR looks way scarier than it is. Probably just implementing `DataTypeCodec::getNamespaceUri` in your DataTypeCodecs.

kevinherron:
Gonna play with this a little more... something about needing the namespace URI on `DataTypeCodec` bothers me.

I need it for `OpcUaXmlEncoder::encodeStruct`:
```java
  @Override
  public void encodeStruct(String field, Object value, DataTypeCodec codec)
      throws UaSerializationException {

    if (beginField(field)) {
      if (depth.getAndIncrement() > context.getEncodingLimits().getMaxRecursionDepth()) {
        throw new UaSerializationException(
            StatusCodes.Bad_EncodingError,
            "max recursion depth exceeded: " + context.getEncodingLimits().getMaxRecursionDepth());
      }

      try {
        String namespaceUri = codec.getNamespaceUri();
        if (namespaceUri == null) {
          throw new UaSerializationException(
              StatusCodes.Bad_EncodingError,
              "no namespace registered for type: " + codec.getEncodingName());
        }

        namespaceStack.push(namespaceUri);
        codec.encode(context, this, value);
        namespaceStack.pop();
      } finally {
        depth.decrementAndGet();
        endField(field);
      }
    }
  }
```

If all the `UaEncoder` and `UaDecoder` encode/decode methods for structs accepted `UaStructuredType` instead of `Object` then I could get it from the type id.

The problem is the legacy BSD codec things don't produce `UaStructuredType` things, they produce the raw-wrapped objects. I have no problem breaking backwards compatibility there and changing that... but I need to figure out if it's even possible.

Sorry, these are notes to self.

PR #1451, opened 2025-04-29
Use namespace index if possible when encoding ExpandedNodeId

### LINKED ISSUE #1450: Use namespace index if possible when encoding ExpandedNodeId
In `OpcUaBinaryEncoder::encodeExpandedNodeId`, if we know the namespace index, then use that when encoding instead of a URI.

It's more compact than encoding a namespace.

It's also a workaround for bugs observed in the CTT and UaExpert, where the Client ignores the explicitly specified namespace URI in the ExpandedNodeId and just uses the namespace index, which is always 0 when the URI is encoded, instead.

kevinherron:
Using the namespace index is more compact than encoding a namespace URI.

It's also a workaround for bugs observed in the CTT and UaExpert, where the Client ignores the explicitly specified namespace URI in the ExpandedNodeId and just uses the namespace index, which is always 0 when the URI is encoded.

fixes #1450

### DIFF
--- a/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/encoding/binary/OpcUaBinaryEncoder.java
+++ b/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/encoding/binary/OpcUaBinaryEncoder.java
@@ -390,6 +390,17 @@ public class OpcUaBinaryEncoder implements UaEncoder {
     NamespaceReference namespace = value.namespace();
     if (namespace instanceof NamespaceUri uri) {
       namespaceUri = uri.namespaceUri();
+
+      // If we know the namespace index, use that for the encoding.
+      // It's more compact than encoding a namespace URI.
+      // It's also a workaround for bugs observed in the CTT and UaExpert, where the Client ignores
+      // the explicitly specified namespace URI in the ExpandedNodeId and just uses the namespace
+      // index, which is always 0 when the URI is encoded.
+      UShort index = context.getNamespaceTable().getIndex(namespaceUri);
+      if (index != null) {
+        namespaceIndex = index.intValue();
+        namespaceUri = null;
+      }
     } else if (value.namespace() instanceof NamespaceIndex index) {
       namespaceIndex = index.namespaceIndex().intValue();
     }
--- a/opc-ua-stack/stack-core/src/test/java/org/eclipse/milo/opcua/stack/core/encoding/binary/ExpandedNodeIdSerializationTest.java
+++ b/opc-ua-stack/stack-core/src/test/java/org/eclipse/milo/opcua/stack/core/encoding/binary/ExpandedNodeIdSerializationTest.java
@@ -15,6 +15,7 @@ import static org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.Unsigned.
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import java.util.UUID;
+import org.eclipse.milo.opcua.stack.core.NamespaceTable;
 import org.eclipse.milo.opcua.stack.core.types.builtin.ByteString;
 import org.eclipse.milo.opcua.stack.core.types.builtin.ExpandedNodeId;
 import org.eclipse.milo.opcua.stack.core.types.builtin.ExpandedNodeId.NamespaceReference;
@@ -53,6 +54,12 @@ public class ExpandedNodeIdSerializationTest extends BinarySerializationFixture 
     writer.encodeExpandedNodeId(nodeId);
     ExpandedNodeId decoded = reader.decodeExpandedNodeId();
 
-    assertEquals(nodeId, decoded);
+    NamespaceTable namespaceTable = writer.getEncodingContext().getNamespaceTable();
+
+    assertEquals(nodeId.getIdentifier(), decoded.getIdentifier());
+    assertEquals(nodeId.getServerIndex(), decoded.getServerIndex());
+    assertEquals(nodeId.getNamespaceUri(namespaceTable), decoded.getNamespaceUri(namespaceTable));
+    assertEquals(
+        nodeId.getNamespaceIndex(namespaceTable), decoded.getNamespaceIndex(namespaceTable));
   }
 }


johannwesely:
Hello, I tried the fix an it works well for me no more errors in CTT Adress Space Base. Thank you very much.

johannwesely:
Hi Kevin,
i updatet to 1.0.0-M2 and realized that this change has been reverted. So I have the Type Cache Error for ObjectTypes again . Why did you revert this ?

Greetings Johann

kevinherron:
I reverted the band-aid fix and replaced it with a fix for the actual root cause: https://github.com/eclipse-milo/milo/commit/4a27357748055561159d08c9550af8f11ba6ffe6

When I tested it with your namespace and the CTT it was still fixed... I'll check again this week I suppose.

kevinherron:
> Make sure that you aren't pulling in a mix of M1 and M2 artifacts. The uanodeset artifact still points to M1 artifacts.

Actually, make sure the version of `uanodeset` you are using is `0.3-M2`, which _does_ point at Milo 1.0.0-M2. Other versions will give you a mismatch.

johannwesely:
Hi,
you are right with ctt the errors are gone but i still see them with UAExpert #1428. Your band-aid fix also worked for UaExpert so i got confused.

johannwesely:
But i sometimes running the testscripts (not always?) get this error in CTT and i think it has to do with the custom object types as well. Address Space Base shows no errors.
Screenshot 2025-05-13 105449

kevinherron:
@johannwesely it seems I had never properly published uanodeset `0.3-M2`. I just re-ran that release and it should be available soon. Make sure that you are using this version, and that you don't have a mix of M1 and M2 artifacts on your class path after you have updated.

Your nodeset file is working in both the CTT and UaExpert without errors for me.

johannwesely:
Hello,
I use uanodeset 0.3-M2 and Milo 1.0.0.-M2  and made sure to not use old dependencies, but still have the same issue. How did you load the Nodeset i sent you for testing. Because there are multiple namespaces inside the .xml file i use and in your example server there is only one ?
 I Also have to do something like this in my DataTypeNamespace  
```
		@Override
		protected boolean filterNamespace(String namespaceUri) {
			return !namespaceUri.equals("http://opcfoundation.org/UA/");
	
		}
```
 otherwise it wont load namespaces at all. Can you explain me what is the diffrence between the quickfix (wich works fine ) and the M2 release for my understanding? Greetings Johann

kevinherron:
> Can you explain me what is the diffrence between the quickfix (wich works fine ) and the M2 release for my understanding? 

The quick fix PR hooked into the binary encoder and replaced _every_ outgoing ExpandedNodeId that had a URI in it with one that used an index instead. The actual fix located the place (browsing) where ExpandedNodeId with a URI was being returned and fixed it, because it turns out [there's a specific passage in the spec](https://reference.opcfoundation.org/Core/Part4/v105/docs/7.30#:~:text=If%20the%20server%20Index%20indicates%20that%20the%20TargetNode%20is%20a%20remote%20Node%2C%20then%20the%20nodeId%20shall%20contain%20the%20absolute%20namespace%20URI.%20If%20the%20TargetNode%20is%20a%20local%20Node%20the%20nodeId%20shall%20contain%20the%20namespace%20index.) that states ExpandedNodeId in the browse results must use an index. 

I think the issue you may be having now is that the type definition in the `ReferenceDescription` is not getting the same treatment, but it seems UaExpert expects it.

kevinherron:
> I Also have to do something like this in my DataTypeNamespace

		@Override
		protected boolean filterNamespace(String namespaceUri) {
			return !namespaceUri.equals("http://opcfoundation.org/UA/");
	
		}

Are you also overriding `getNodeSetInputStreams` and returning a List of all the NodeSet XML files?

kevinherron:
@johannwesely I need to write up an issue and talk with some OPC peeps, but a patch for Milo that I think fixes your issue:

```
Subject: [PATCH] Convert type definitions to relative ExpandedNodeId
---
Index: opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/servicesets/impl/helpers/BrowseHelper.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/servicesets/impl/helpers/BrowseHelper.java b/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/servicesets/impl/helpers/BrowseHelper.java
--- a/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/servicesets/impl/helpers/BrowseHelper.java	(revision e921de7336c92c127b4bb1398b07d66af8832740)
+++ b/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/servicesets/impl/helpers/BrowseHelper.java	(date 1747745278911)
@@ -369,7 +369,8 @@
     var typeDefinitions = new ArrayList<ExpandedNodeId>();
 
     for (int i = 0; i < nodeIds.size(); i++) {
-      NodeId nodeId = nodeIds.get(i).toNodeId(server.getNamespaceTable()).orElse(NodeId.NULL_VALUE);
+      NamespaceTable namespaceTable = server.getNamespaceTable();
+      NodeId nodeId = nodeIds.get(i).toNodeId(namespaceTable).orElse(NodeId.NULL_VALUE);
 
       BrowseAttributes attributes = browseAttributes.get(i);
       NodeClass nodeClass = attributes.nodeClass;
@@ -380,7 +381,17 @@
         switch (attributes.nodeClass) {
           case Object, Variable -> {
             try {
-              typeDefinitions.add(browseTypeDefinition(server, nodeId));
+              ExpandedNodeId typeDefinitionId = browseTypeDefinition(server, nodeId);
+              if (typeDefinitionId.isLocal()) {
+                if (typeDefinitionId.isAbsolute()) {
+                  typeDefinitionId = typeDefinitionId.relative(namespaceTable).orElseThrow();
+                }
+              } else {
+                if (typeDefinitionId.isRelative()) {
+                  typeDefinitionId = typeDefinitionId.absolute(namespaceTable).orElseThrow();
+                }
+              }
+              typeDefinitions.add(typeDefinitionId);
             } catch (UaException e) {
               LoggerFactory.getLogger(BrowseHelper.class)
                   .error("Error browsing TypeDefinition for nodeId={}", nodeId, e);
```

It basically just applies the same treatment to the type definition returned in a ReferenceDescription as the target NodeId gets.

johannwesely:
I apllied your patch and it works. Thank you . So will this be in the release also?

kevinherron:
1.0.0 was released yesterday, so this will have to wait a bit for the 1.0.1 patch release.

PR #1463, opened 2025-05-19
add encodeBuiltinType and encodeBuiltinTypeArray to UaEncoder interface

vpriscan:
- added common encodeBuiltinType and encodeBuiltinTypeArray to UaEncoder interface

- aligning implementations in OpcUaJsonEncoder, OpcUaXmlEncoder and OpcUaBinaryEncoder


vpriscan:
@kevinherron 
In our software that uses Milo, we have _value:Object_ and _dataType:OpcUaDataType_ variables.
We want to encode this value with UaEncoder (json, xml or binary),
but encodeBuiltin* functions that accept **Object** value are all private in UaEncoder implementations.
To avoid having to re-implement those functions, it would be highly useful to make them public and move them to UaEncoder interface.

Equivalent would also be useful for UaDecoder as well.

PR #1501, opened 2025-06-25
Provide a "bill of materials" (BOM) module

comtel2000:
Here is a small feature request to provide an easier version check for projects that uses multiple Milo modules.

It adds a BOM (Bill of Materials) module in order to avoid different versions and allow projects a central version control:

```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>org.eclipse.milo</groupId>
      <artifactId>milo-bom</artifactId>
      <version>${milo.version}</version>
      <type>pom</type>
      <scope>import</scope>
     </dependency>
   </dependencies>
</dependencyManagement>
```

I had to make adjustments to these WIP sub modules (json-codec/https/websocket) because the explicit versioning causes issues.

Is the special versioning (0.1-SNAPSHOT) intended as a release filter?


kevinherron:
> Is the special versioning (0.1-SNAPSHOT) intended as a release filter?

It's not so much a release filter as it is an explicit indication that these are incubating / WIP code. They are not released, but the mechanism that makes modules I don't want to release happen is mostly the deploy plugin being disabled in the corresponding `pom.xml`:
```
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-deploy-plugin</artifactId>
        <configuration>
          <skip>true</skip>
        </configuration>
      </plugin>
```

kevinherron:
You mean another maven module or something else?

I've never really minded explicit versioning. I think I prefer it most of the time. Versions are usually either pointing to a property, so they can be updated in one place, or if they're the project/module version they are updated by the Maven versions plugin when preparing a release.

comtel2000:
I used this pattern:
https://github.com/anliksim/maven-template-bom

which does not work well with explicit sub versions.
I will try to find a way without manual version synchronization.

comtel2000:
Yes it's without the "eat your own dog food" part :)

kevinherron:
Maybe we can eat the dogfood in another project. https://github.com/digitalpetri/opc-ua-demo-server?

It only has one Milo dependency for `milo-sdk-server`, but still something...

comtel2000:
No, I guess it's ready. I also started a dryRun with the release plugin and all versions and modules could be created.

kevinherron:
With the move to GH actions the release plugin is no longer used for release.

The process to release 1.0.2 would be something like:
- create a `release/1.0.2` branch
- set the version to `1.0.2` via `mvn versions:set`
- commit, push the commit
- create a GH release from this commit, creating a `v1.0.2` tag
- this triggers the [release workflow](https://github.com/eclipse-milo/milo/blob/1.0/.github/workflows/maven-release.yml), which runs a `deploy`, which is picked up by the new `central-publishing-maven-plugin`

It took a lot of build hacking to get that stupid plugin to not publish the test and incubating artifacts...

PR #1637, opened 2025-10-30
Fix SessionFuture completion and ChannelFsm listener lifecycle bugs

kevinherron:
Complete pending SessionFuture instances when session creation is aborted via CloseSession events. The CreatingWait → Inactive and ReactivatingWait → Closing transitions now call handleFailureToOpenSession() to ensure any GetSession or OpenSession futures are properly completed with Bad_SessionClosed exceptions.

Prevent ChannelFsm listener accumulation when cycling through Active state. Multiple listeners could accumulate if keep-alive timeouts triggered reactivation without actual channel disconnection. Listeners are now stored in FSM context and explicitly removed when transitioning away from Active state, ensuring exactly one listener is registered during each Active period.


ia-cmorgan (approved):
LGTM

PR #1726, opened 2026-04-03
Decouple OpcUaSubscription.SyncState from MonitoredItem operations

kevinherron:
SyncState now tracks only subscription settings synchronization. MonitoredItem add/remove no longer sets UNSYNCHRONIZED, and create() always sets SYNCHRONIZED. New isMonitoredItemsSynchronized() and isFullySynchronized() methods provide item-level and aggregate sync checks.


kevinherron:
@copilot run `mvn spotless:apply` and then commit+PR the formatting changes it produces.

PR #1757, opened 2026-05-29
Fix OptionSet variant JSON encoding for empty arrays and scalars

LivingLikeKrillin:
- [x] You have signed an Eclipse Contributor Agreement and are committing using the same email address
- [x] Code follows the style guidelines and Checkstyle passes

The `OPTION_SET` branch in `OpcUaJsonEncoder.encodeVariantValue` derived the JSON `typeId` by calling `Array.get(value, 0)` on the variant value. This failed in two real cases: empty arrays threw `ArrayIndexOutOfBoundsException`, and single (non-array) `OptionSet` values threw `IllegalArgumentException("Argument is not an array")`. The latter is reachable via the existing scalar pattern in production code, e.g. `new Variant(defaultAccessRestrictions)` in `NamespaceMetadataTypeNode`.

Derive the `typeId` from the value's class instead, using the existing `OptionSetUI8`/`UI16`/`UI32`/`UI64` subclasses — the approach the original TODO comment in this branch proposed. `OptionSetUI64` is supported preemptively even though no DataType in stack-core currently extends it.

Test coverage in `OpcUaJsonEncoderTest.encodeVariantOptionSet` covers the width × cardinality matrix for the OptionSet subclasses in current use: scalar UI8, non-empty UI8/UI16/UI32, and empty UI8/UI16/UI32.


kevinherron:
Thank you for contributing.

I've already fixed this as part of a larger batch of JSON encoding fixes, and your PR just reminded me that I need to push/PR/merge them :)

See https://github.com/eclipse-milo/milo/pull/1758

So I will close this, but thank you anyway.

PR #1772, opened 2026-06-10
server-examples: add missing DataTypeEncoding Nodes for static custom types

### LINKED ISSUE #1771: server-examples: registerCustomStructType/registerCustomUnionType don't create the DataTypeEncoding node — strict clients can't decode the values
Hi! While digging into why node-opcua only partially decodes custom structs from Milo-based servers (related, but a different cause: digitalpetri/uanodeset#14), we noticed that in the example server (`ExampleNamespace`, v1.1.4) `registerCustomStructType()` and `registerCustomUnionType()` set the `DataTypeDefinition` and register the codec, but don't create the `DataTypeEncoding` node ("Default Binary") or the `HasEncoding` reference — so the encoding NodeId that goes out with every `ExtensionObject` (`ns=2;s=DataType.CustomStructType.BinaryEncoding`) doesn't exist in the address space.

`registerDynamicStructType()` in the same file does the full modeling, which makes for a neat A/B comparison on an unmodified v1.1.4 example server:

- `HelloWorld/CustomStructTypeVariable` → node-opcua: `Cannot find DataType for binary encoding ns=2;s=DataType.CustomStructType.BinaryEncoding`, value stays opaque
- `HelloWorld/DynamicStructTypeVariable` → decodes fine

Clients that resolve encodings via the inverse `HasEncoding` reference (node-opcua) hit this; clients that use `defaultEncodingId` from the `DataTypeDefinition` (asyncua, UaExpert) don't notice, so it's easy to miss. Since the example is what people build their own namespaces from (that's how we ended up debugging it — a server modeled on `registerCustomStructType` showed exactly this), it seems worth completing the static samples too.

If it's of interest: it looks like the encoding nodes used to come for free from the `DataTypeDictionaryManager` in 0.6.x, and quietly disappeared from these samples when the dictionary machinery was removed in 1.0.

Happy to open a small PR adding the encoding node + `HasEncoding` block (mirroring `registerDynamicStructType`) to both methods — and curious whether you'd consider this a bug in the samples or intended minimalism.

mschm:
Fixes #1771.

`registerCustomStructType()` and `registerCustomUnionType()` now create the "Default Binary" `DataTypeEncoding` Node and the `HasEncoding` Reference, like `registerDynamicStructType()` already did. The encoding node creation is extracted into a shared helper used by all three.

Verified against an unmodified node-opcua 2.173.x client: before the change, `HelloWorld/CustomStructTypeVariable` and `HelloWorld/CustomUnionTypeVariable` stay opaque (`Cannot find DataType for binary encoding ns=2;s=DataType.CustomStructType.BinaryEncoding`); with it, all three custom type variables decode.

### DIFF
--- a/milo-examples/server-examples/src/main/java/org/eclipse/milo/examples/server/ExampleNamespace.java
+++ b/milo-examples/server-examples/src/main/java/org/eclipse/milo/examples/server/ExampleNamespace.java
@@ -805,6 +805,9 @@ public class ExampleNamespace extends ManagedNamespaceWithLifecycle {
 
     getNodeManager().addNode(dataTypeNode);
 
+    // Add a DataTypeEncodingNode for binary encoding of the new DataType
+    addDataTypeEncodingNode(dataTypeId, binaryEncodingId);
+
     // Define the structure
     StructureField[] fields =
         new StructureField[] {
@@ -880,6 +883,9 @@ public class ExampleNamespace extends ManagedNamespaceWithLifecycle {
 
     getNodeManager().addNode(dataTypeNode);
 
+    // Add a DataTypeEncodingNode for binary encoding of the new DataType
+    addDataTypeEncodingNode(dataTypeId, binaryEncodingId);
+
     StructureField[] fields =
         new StructureField[] {
           new StructureField(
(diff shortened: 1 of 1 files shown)


kevinherron:
@mschm This looks good, but you must have an [ECA](https://www.eclipse.org/legal/eca/) signed and on file with the Eclipse Foundation before I'm allowed to accept contributions.

If you don't want to bother I can just apply this myself in a different PR/commit.

mschm:
@kevinherron ECA is now in place

kevinherron:
@mschm thanks again :)

PR #1862, opened 2026-09-03
Make CertificateGroup the unit of identity and trust

### LINKED ISSUE #1861: Remove CertificateFactory from CertificateGroup; make provisioning a factory-side helper
## Problem

`CertificateGroup` exposes `getCertificateFactory()`, and `DefaultApplicationGroup` takes a `CertificateFactory` in every constructor. The only thing the group ever does with it is generate self-signed material for supported certificate types the store is missing (`createMissingCertificates()`, formerly `initialize()`).

A certificate group in the Part 12 sense is a set of certificate types, a trust list, and the validation policy that goes with it. The store, trust list manager, and validator fit that model. Creating keys and certificates does not; it is an application deployment choice (self-signed bootstrap vs. GDS-issued vs. pre-provisioned) that the group has no reason to own.

## History

The factory has been on the group since the CertificateGroups refactor (#1154). The stated motivation was the GDS push model: a server answering `CreateSigningRequest` for a group needs to generate a key pair and CSR for one of that group's certificate types, and hanging the factory off the group made the lookup trivial. The push model server side was never implemented in Milo, so across the whole history the only production consumer of `getCertificateFactory()` has been the group's own bootstrap loop. The factory was designed in anticipation of a consumer that never materialized.

## Proposal

- Drop `getCertificateFactory()` from `CertificateGroup`.
- Drop the `CertificateFactory` parameter from `DefaultApplicationGroup` constructors. Keep the existing constructors as deprecated overloads for compatibility.
- Move provisioning to the factory side, written against the `CertificateGroup` interface so it works for any group implementation:

  ```java
  // CertificateFactory
  default List<NodeId> createMissingCertificates(CertificateGroup group) throws Exception
  ```

  Call sites become:

  ```java
  var group = new DefaultApplicationGroup(trustListManager, certificateStore, certificateValidator);
  certificateFactory.createMissingCertificates(group);
  ```

- The deprecated `createAndInitialize(...)` statics already accept a factory; they delegate to the two-step form above and keep pointing callers at it.

## Open questions

1. **Presence vs. readability.** The current loop uses `CertificateStore.contains(typeId)`, which distinguishes an absent entry from an unreadable one. A helper written against `CertificateGroup` would judge "missing" by `getCertificateChain(typeId)` returning empty, and the group returns empty for read failures too, so a corrupt entry would be silently replaced. Options: add a `hasCertificate(typeId)` presence check to the interface, or document the helper as replacing anything it cannot read. Leaning toward the explicit presence check.
2. **Does push support need a per-group factory?** Part 12 lets each group advertise its own certificate types, so different groups could want different issuance policy. When Milo gains push support, a group-id-keyed factory map on the server configuration should cover this without the group holding a factory. Worth confirming against a real push implementation before committing to the shape.
3. **Does `createCertificateChain` belong on `CertificateFactory` at all?** Key pair and CSR creation are what GDS flows need. Self-signed chain creation is bootstrap-only and `SelfSignedCertificateBuilder` already covers it directly.

## Before changing

Survey downstream consumers that build on these APIs, in particular ones that implement GDS push on top of Milo, and check:

- [ ] How they construct `DefaultApplicationGroup` / implement `CertificateGroup`, and whether any subclass or wrapper relies on the factory constructor parameter.
- [ ] Whether their push implementation obtains the factory via `CertificateGroup.getCertificateFactory()` for `CreateSigningRequest` / `UpdateCertificate` (regenerate-private-key case), or sources it elsewhere.
- [ ] Whether they need per-group issuance policy (question 2) or a single per-type factory is enough.
- [ ] How disruptive the deprecated-constructor path is for them versus a clean break.

## Context

Follow-up to #1855. The intermediate step, dropping the initialization lifecycle and renaming `initialize()` to `createMissingCertificates()`, is already on that branch.

kevinherron:
This makes `CertificateGroup` the complete unit of local identity and trust on both the client and the server. Groups carry no NodeId and no `CertificateFactory`. Servers register groups under `CertificateGroupType` NodeIds in the `CertificateManager`, clients configure one group (or one key pair and certificate), and self-signed provisioning is a factory-side helper. These are breaking changes on `integration/1.2` with no deprecation shims.

### Issue #1861: remove the factory from the group

`CertificateGroup.getCertificateFactory()` is gone, and `DefaultCertificateGroup` (renamed from `DefaultApplicationGroup`) no longer takes a factory. Provisioning is `CertificateFactory.createMissingCertificates(CertificateGroup)`, written against the interface so it works for any implementation. It uses a new `CertificateGroup.hasCertificate(NodeId)` presence check, which resolves the issue's first open question: an entry the store cannot read is reported, never silently replaced with a fresh self-signed identity. The issue's second question (per-group factory for push) is left for a push implementation; the NodeId registry below is the seam it would resolve groups through. `createCertificateChain` stays on the factory (third question) because the provisioning helper needs it.

### Beyond the issue

Working through #1861 against the Ignition GDS branch showed that the group's NodeId and the client's use of `CertificateManager` were the same design problem, so this PR takes them on together.

- **Groups have no NodeId.** `getCertificateGroupId()` is gone from the interface and `Entry` is `(certificateTypeId, certificateChain)`. `CertificateIdentity` references its `CertificateGroup`. A client no longer mints NodeIds it has no namespace for.
- **`CertificateManager` is the server-side registry.** `DefaultCertificateManager(defaultGroup)` registers the group under `DefaultApplicationGroup`; `addCertificateGroup(nodeId, group)` adds more. Registration order defines cross-group selection precedence, replacing a sort on the NodeId string. `getCertificateGroupId(group)` serves code that builds address space nodes. `OpcUaServer` omits an endpoint whose group is not registered, with reason `certificate group not registered`.
- **Identity selection takes candidate groups.** `CertificateIdentitySelectionContext` carries a `List<CertificateGroup>` instead of a manager and group id. The explicit certificate pin remains for the server's per-endpoint certificate only.
- **The quarantine belongs to the group.** `CertificateManager.getCertificateQuarantine()` is gone; `CertificateGroup.getCertificateQuarantine()` replaces it. `CertificateManager.getRejectedCertificates()` is the server-wide list for `GetRejectedList`: the union over `getCertificateGroups()` in registration order, de-duplicated at first occurrence.
- **The client configures one group.** `OpcUaClientConfigBuilder.setCertificateGroup` replaces `setCertificateManager`, `setCertificateGroupId`, `setKeyPair`, `setCertificate`, and `setCertificateChain`. The validator defaults to the group's. `ClientApplicationContext` exposes only `getCertificateIdentity` and `getCertificateValidator`; a secured endpoint with no identity fails with `Bad_ConfigurationError` rather than falling back to fixed key material. The advertised ApplicationUri comes from the selected identity, or on `None` from the URI shared by the group's identities.
- **Simple clients keep a simple path.** `setCertificateIdentity(keyPair, certificateChain)` wraps a key pair and chain in an internal group of one. The client never reads that group's trust material; servers are validated through the configured validator as before. `DefaultCertificateGroup.forIdentity` is the public group-of-one constructor for callers that need real trust material on the group, such as the GDS pull example.

### Notes for downstream implementers

- Custom `CertificateGroup` implementations drop `getCertificateGroupId()` and `getCertificateFactory()` and add `hasCertificate(NodeId)` and `getCertificateQuarantine()`.
- Custom `CertificateManager` implementations add `getCertificateGroupId(CertificateGroup)` and drop the quarantine. `getCertificateIdentities()` now returns registration order across groups.
- A push implementation that read the factory from the group keeps its own factory reference.
- `OpcUaClientConfig.copy` now copies the validator; previously a copy fell back to the insecure default.
- `DefaultCertificateManager.getCertificateGroupId(group)` matches by instance. A delegating wrapper registered by its delegate is not found.
- A `setCertificate` pin migrated to `setCertificateIdentity` is now a group of one with an inferred certificate type, and the type gates which security policies it is offered for. A SHA-1 signed RSA certificate is RsaMin, so it fails client-side with `Bad_ConfigurationError` on a Basic256Sha256 endpoint instead of being sent and rejected by the server. A key type that cannot be classified fails at `build()`. The error message now says whether no group is configured or no configured identity is compatible.
- `CertificateIdentity` equality includes the group reference, so two reads of the same store-backed group yield non-equal identities. Compare by group, type, and certificate where value equality is needed.

Verification:

- `mise exec -- mvn -q spotless:apply`
- `mise exec -- mvn -q clean compile`
- `mise exec -- mvn -q clean verify`
- `mise exec -- mvn -q install -DskipTests`

Closes #1861


kevinherron:
Adopted this on Ignition's GDS branch today; it lands cleanly. The client-side `CertificateManager`, the URN-prefixed ns=0 group id, and the fixed-certificate pin are all gone, and the test servers lost their placeholder factories. Full module suite and the Docker integration test are green. A few notes from the adoption:

1. GetRejectedList aggregation is now every server's job. A default `CertificateManager.getRejectedCertificates()` returning the de-duplicated union over `getCertificateGroups()` would save each consumer writing the same loop and choosing a de-dup rule.
2. The explicit-alias path behaves differently in two edge cases once it becomes a group of one: a SHA-1 signed RSA certificate is inferred as RsaMin and now fails client-side with `Bad_ConfigurationError` on a Basic256Sha256 endpoint instead of being sent and rejected by the server, and `forIdentity` throws for a key type it cannot classify. Both are acceptable to me, but they are behavior changes for anyone migrating a `setCertificate` pin.
3. Minor: `DefaultCertificateManager.getCertificateGroupId(group)` compares by reference. Fine, but the Javadoc could say "the same instance", since a delegating wrapper registered by its delegate is not found.

Nothing blocking from my side.

PR #1864, opened 2026-09-05
Preserve server startup cleanup and shutdown ordering

kevinherron:
This PR fixes two server lifecycle failures that can leave namespace resources registered or tear them down while startup is still running.

### Roll back children after a failed startup

LifecycleManager previously let a child startup failure escape without stopping earlier children. It then became stopped and ignored subsequent shutdown, which could leave an address space and node manager registered.

It now unwinds successfully started children in reverse order, continues cleanup after individual failures, and attaches those failures to the original startup exception. The child whose own startup fails remains responsible for its partial initialization.

### Keep public future completion separate from shutdown ordering

Concurrent callers could cancel or complete the shared server startup future while a participant callback was still running. Shutdown treated that future as proof that startup had finished and removed core facilities before participant cleanup.

A private completion future now controls shutdown ordering, while callers still share the same public result. Changing that result cannot release the internal wait. Direct completion preserves the original startup exception for callback users.

### Verification

Regression tests cover child rollback under both shutdown-order configurations, preserved cleanup exceptions, real namespace registration cleanup, and premature shutdown after cancellation or external completion of the public future. They produced six assertion failures against the original production code.

Formatting, a full clean compile, and selected lifecycle tests passed. Verification used `mise exec -- mvn -q spotless:apply`, `mise exec -- mvn -q clean compile`, and SDK server reactor `verify` covering LifecycleManagerTest, AbstractLifecycleTest, and OpcUaServerLifecycleParticipantTest.

### DIFF
--- a/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/LifecycleManager.java
+++ b/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/LifecycleManager.java
@@ -11,6 +11,8 @@
 package org.eclipse.milo.opcua.sdk.server;
 
 import com.google.common.collect.Lists;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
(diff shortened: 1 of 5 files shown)


kevinherron:

- [Roll back started children after lifecycle startup fails](https://github.com/eclipse-milo/milo/pull/1876)
- [Keep server startup completion private from callers](https://github.com/eclipse-milo/milo/pull/1879)

PR #1865, opened 2026-09-05
Keep subscription recovery and delayed responses with their lifecycle

kevinherron:
This PR addresses six client recovery and concurrency issues that can stop publishing, apply stale responses, or stall lifecycle transitions. It is stacked on #1863, which supplies the monitored-item response guards tested here.

### Preserve session callback order during reactivation

A delayed inactive notification could arrive after the same Session reactivated and close its recovered publishing pipeline. Session activity callbacks now execute in order through a sequential executor.

### Reject Publish responses for a replaced registration

A pending Publish response could resolve a reused subscription ID to a replacement registration. Each request now captures the registration map, and response processing checks the original registration's identity before applying the result.

Subscriptions first registered after dispatch and unrelated registration changes remain valid. [Part 4 §5.14.5.1](https://reference.opcfoundation.org/Core/Part4/v105/docs/5.14.5) explains why: "Since Publish requests are not directed to a specific Subscription, they may be used by any Subscription." The wire identifier does not distinguish successive Java registrations.

### Synchronize reset with monitored-item changes

Reset could race with monitored-item add/remove and leave a retained item without its client handle. Reset now uses the same collection lock as add/remove, in the established lifecycle-to-items lock order.

### Protect recreated items from delayed service responses

New concurrent regressions cover delayed CreateMonitoredItems, ModifyMonitoredItems, DeleteMonitoredItems, and SetMonitoringMode responses after reset and recreation. The production guards are already in #1863; these tests verify that each rejects a stale response and preserves the replacement item's state.

[Part 4 §5.13.2.2](https://reference.opcfoundation.org/specs/OPC-10000-4/5.13.2.2) scopes monitored-item identifiers: "This id is unique within the Subscription, but might not be unique within the Server or Session."

### Complete lifecycle transitions when an executor rejects work

Rejected completion handoffs could strand queued subscription operations. Direct execution could also recurse through enough handoffs to stall the queue. A per-subscription sequential executor now drains transitions iteratively, with a caller-thread fallback when dispatch is rejected.

### Invalidate operation limits without blocking session recovery

Transfer cleanup could wait for an operation-limits Read whose completion depended on session activation finishing. The cache now uses NonBlockingLazy, so reset invalidates an in-progress generation without waiting for its supplier.

### Verification

Formatting, a full clean compile, and selected session, subscription, monitored-item, and publishing tests passed. The new tests reproduce the defects present on the stacked base. Removing the four existing monitored-item response guards in an isolated baseline makes all four corresponding regressions fail. Transition tests cover 10,000 queued operations under both rejected and direct execution.

### DIFF
--- a/opc-ua-sdk/integration-tests/src/test/java/org/eclipse/milo/opcua/sdk/client/session/SessionActivityOrderingTest.java
+++ b/opc-ua-sdk/integration-tests/src/test/java/org/eclipse/milo/opcua/sdk/client/session/SessionActivityOrderingTest.java
@@ -1,0 +1,271 @@
+/*
+ * Copyright (c) 2026 the Eclipse Milo Authors
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.milo.opcua.sdk.client.session;
+
+import static org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.Unsigned.ubyte;
+import static org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.Unsigned.uint;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+import com.digitalpetri.fsm.Fsm;
+import io.netty.util.HashedWheelTimer;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayDeque;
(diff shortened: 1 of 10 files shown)


kevinherron:

- [Keep Session activity callbacks in lifecycle order](https://github.com/eclipse-milo/milo/pull/1901)
- [Keep delayed Publish responses with their registration](https://github.com/eclipse-milo/milo/pull/1904)
- [Coordinate monitored-item reset with add and remove](https://github.com/eclipse-milo/milo/pull/1907)
- [Cover delayed monitored-item responses after recreation](https://github.com/eclipse-milo/milo/pull/1910)
- [Drain subscription transitions after executor rejection](https://github.com/eclipse-milo/milo/pull/1913)
- [Keep subscription reset independent of limits lookup](https://github.com/eclipse-milo/milo/pull/1916)

PR #1866, opened 2026-09-05
Preserve dynamic values and complete type discovery correctly

kevinherron:
This PR fixes four serialization and type-discovery issues that can lose values, break round trips, or prevent type-tree construction.

### Keep optional-field mask positions stable

When an earlier optional field was absent, both dynamic structure codecs assigned the next present field the wrong mask bit. The binary codec silently wrote the wrong field; the JSON-backed codec could throw while reading an absent member. Both now advance the optional index for every declared optional field. Part 6 §5.2.7 assigns positions independently of presence: "The first optional field is assigned bit '0', the second optional field is assigned bit '1'". Its worked example also uses mask `0x02` when only the second optional field is present. [OPC 10000-6 §5.2.7](https://reference.opcfoundation.org/specs/OPC-10000-6/5.2.7).

### Decode XML Variants with whitespace and comments

XML Variant decoding now skips whitespace and comments between elements, and uses the existing matrix decoder to validate shapes and element types. Previously, indentation could turn an Int32 Variant into null or make matrix traversal read the wrong nodes. Part 6 §5.3.1.17 specifies that a scalar Variant "shall contain a single child element with the name of the built-in type"; the same clause defines the Matrix Dimensions/Elements layout. [OPC 10000-6 §5.3.1.17](https://reference.opcfoundation.org/specs/OPC-10000-6/5.3.1.17).

### Retain fallback encoding IDs for decoded structures

When a server omits HasEncoding references, ClientDataType now retains the StructureDefinition fallback encoding ID. Previously, registration used the fallback but decoded values reported a null encoding ID, so re-encoding failed. Explicit encoding references still take precedence, and absent encoding IDs remain null for abstract or nested-only structures. Part 3 §8.48 specifies, after its abstract/nested-type exceptions: "In all other cases the DefaultEncodingId in the StructureDefinition shall be the Default Binary encoding for the DataType". [OPC 10000-3 §8.48](https://reference.opcfoundation.org/specs/OPC-10000-3/8.48).

### Stop browsing when the continuation token is empty

Type-tree browsing now stops for an empty continuation token as well as a null token. A completed Browse previously caused an unnecessary BrowseNext and could repeat until tree construction failed. Part 4 §7.9 explains the continuation contract: "The Server allocates a ContinuationPoint if there are more results to return." The fix also follows the helper's existing null-or-empty terminal contract. [OPC 10000-4 §7.9](https://reference.opcfoundation.org/specs/OPC-10000-4/7.9).

### Verification

Regression tests failed against the original production code for all four defects. They cover sparse optional scalar and array fields against an independent encoder, compact versus indented/commented XML, malformed matrix shapes, structure and union decode/re-encode with fallback or explicit IDs, and initial/final empty continuation tokens.

Verification passed: `mise exec -- mvn -q spotless:apply`, `mise exec -- mvn -q clean compile`, and targeted reactor `verify` for dynamic serialization, JSON codecs, XML codecs, and client type-tree tests (zero failures, errors, or skips). No foreign-server interoperability run was performed.

Existing empty-matrix and BrowseResult status-handling limitations are unchanged.

### DIFF
--- a/opc-ua-sdk/codec-json/src/main/java/org/eclipse/milo/sdk/core/types/json/JsonStructCodec.java
+++ b/opc-ua-sdk/codec-json/src/main/java/org/eclipse/milo/sdk/core/types/json/JsonStructCodec.java
@@ -606,8 +606,11 @@ public class JsonStructCodec extends GenericDataTypeCodec<JsonStruct> {
     if (definition.getStructureType() == StructureType.StructureWithOptionalFields) {
       int optionalFieldIndex = 0;
       for (StructureField field : fields) {
-        if (field.getIsOptional() && value.getJsonObject().has(requireNonNull(field.getName()))) {
-          switchField = switchField | (1L << optionalFieldIndex++);
+        if (field.getIsOptional()) {
+          if (value.getJsonObject().has(requireNonNull(field.getName()))) {
+            switchField |= 1L << optionalFieldIndex;
+          }
+          optionalFieldIndex++;
         }
       }
       encoder.encodeUInt32("SwitchField", UInteger.valueOf(switchField));
--- a/opc-ua-sdk/codec-json/src/main/java/org/eclipse/milo/sdk/core/types/json/package-info.java
+++ b/opc-ua-sdk/codec-json/src/main/java/org/eclipse/milo/sdk/core/types/json/package-info.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2026 the Eclipse Milo Authors
+ *
+ * This program and the accompanying materials are made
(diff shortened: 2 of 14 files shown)


kevinherron:

- [Preserve mask positions when optional structure fields are absent](https://github.com/eclipse-milo/milo/pull/1882)
- [Decode XML Variants independently of indentation](https://github.com/eclipse-milo/milo/pull/1885)
- [Stop type discovery when the Browse continuation token is empty](https://github.com/eclipse-milo/milo/pull/1888)
- [Retain a structure definition's default binary encoding](https://github.com/eclipse-milo/milo/pull/1891)

PR #1867, opened 2026-09-05
Fix trust-list removal persistence and keystore file watching

kevinherron:
This PR fixes two filesystem persistence problems in certificate management.

### Remove imported trust material durably

Removing imported trust material could leave its original file on disk, causing the entry to return after reload or restart. The trust-list manager now removes matching material from every accepted file, including arbitrary filenames and duplicates. Removing a CRL from a bundle preserves its unrelated entries and the original file's POSIX permissions. A rewritten CRL bundle uses concatenated DER, including when the original file contained PEM.

[Part 12 §7.8.2.7](https://reference.opcfoundation.org/specs/OPC-10000-12/7.8.2.7) describes the removal operation: "The RemoveCertificate Method allows a Client to remove a single Certificate from the TrustList." Disk durability follows Milo's file-backed trust-list contract; the specification does not prescribe filenames or a storage algorithm. File operations retain existing best-effort error handling.

### Follow keystore symlinks when watching for changes

The watcher previously watched only the configured file's directory, so replacement of a target in another directory could go unnoticed. It now follows both the configured path and its resolved symlink target, reloads after target-file replacement, follows link retargeting, and preserves a shared directory watch when both paths use the same directory. Closing the watcher during retargeting exits normally.

Intermediate symlink-chain changes and a newly dangling target in an unwatched directory remain outside this change.

### Verification

Formatting, a full clean compile, and the full stack-core suite passed. Real filesystem regressions cover close/reopen persistence, duplicate imports, multi-CRL bundles, target rotation, and retargeting. All eight new cases failed on the base branch and pass with the fixes; none were skipped on this Linux run. After the permission-preservation and watcher-close changes, formatting, a full clean compile, and selected certificate tests passed with no failures, errors, or skips.

### DIFF
--- a/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/security/FileBasedTrustListManager.java
+++ b/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/security/FileBasedTrustListManager.java
@@ -469,6 +469,9 @@ public class FileBasedTrustListManager implements TrustListManager, Closeable {
     // failure cannot destroy the unrelated CRLs that it also contains.
     Path temporary = Files.createTempFile(file.getParent(), ".crl-", ".tmp");
     try {
+      if (file.getFileSystem().supportedFileAttributeViews().contains("posix")) {
+        Files.setPosixFilePermissions(temporary, Files.getPosixFilePermissions(file));
+      }
       try (var output = Files.newOutputStream(temporary)) {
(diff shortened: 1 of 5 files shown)


kevinherron:

- [Remove imported trust material from its actual files](https://github.com/eclipse-milo/milo/pull/1911)
- [Observe keystore symlinks and their resolved targets](https://github.com/eclipse-milo/milo/pull/1914)

PR #1868, opened 2026-09-05
Fix node instantiation ownership, model freshness, and diagnostics IDs

kevinherron:
This PR fixes seven issues in node instantiation, cleanup, and diagnostics child allocation.

### Preserve replacement nodes during cleanup

Cleanup through an old instantiation result could remove a replacement node or reference with the same logical identity. Cleanup and rollback now remove entries only while they retain the recorded object identities. Custom NodeManager implementations must preserve reference object identity or override conditional removal for their storage model.

### Avoid lock inversion during node-manager callbacks

Reading node attributes or invoking reference predicates while holding the manager monitor could deadlock with attribute observers. Node identity reads and reference predicates now run outside that monitor.

### Retire pending cache loads during invalidation

A load started before invalidation could publish a stale compiled model afterward. Invalidation now retires pending loads, and callers retry against the current cache generation.

### Distinguish values in revision fingerprints

Ambiguous value concatenation could assign the same revision fingerprint to different declaration attributes. Fingerprints now encode type, length, and element boundaries. Binary values use compact hex encoding to retain that distinction without excessive expansion.

### Snapshot mutable declaration values

Compiled plans could retain mutable nested declaration values and change after compilation. Snapshots now copy nested standard OPC UA structures and arrays, including empty matrices without type metadata. Custom Java values that lack a usable standard codec remain opaque and must be immutable.

Part 3 §6.4.2 says, "If new copies are created, then the Attribute values of the InstanceDeclarations are used as the initial values." Snapshot isolation and cache freshness preserve the declaration values used by the selected plan. [Creating an instance](https://reference.opcfoundation.org/specs/OPC-10000-3/6.4.2).

### Resolve references through overridden declarations

References to an overridden declaration could resolve back into type space instead of to the effective instance occurrence. The compiler now maps ancestor declaration identities to that occurrence.

Part 3 §6.3.3.3 says, "A subtype overrides an InstanceDeclaration by specifying an InstanceDeclaration with the same BrowsePath." Replication of nonhierarchical references remains server-specific; this fix makes the references Milo chooses to copy resolve consistently. [Overriding InstanceDeclarations](https://reference.opcfoundation.org/specs/OPC-10000-3/6.3.3.3).

### Allocate diagnostics children independently of array size

Diagnostics arrays used live list size as a child identity, which could collide after an earlier entry was removed and prevent insertion. Subscription, session, and security diagnostics arrays now allocate monotonic child identities.

Part 5 §7.11 says, "For each entry of the array, instances of this type will provide a Variable of the SubscriptionDiagnosticsType VariableType". Distinct identities preserve that behavior as subscriptions enter and leave. Individual session and security diagnostics children remain optional. [SubscriptionDiagnosticsArrayType](https://reference.opcfoundation.org/specs/OPC-10000-5/7.11).

### Verification

The original 17 regression cases failed against the base production code. Formatting, a full clean compile, and unfiltered SDK server/dependency verification passed with zero failures or errors and two skips.

Four additional compatibility and fingerprint regressions failed against the intermediate implementation. After their corrections, formatting, a full clean compile, and selected tests passed with no failures, errors, or skips.

### DIFF
--- a/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/AbstractNodeManager.java
+++ b/opc-ua-sdk/sdk-server/src/main/java/org/eclipse/milo/opcua/sdk/server/AbstractNodeManager.java
@@ -13,7 +13,9 @@ package org.eclipse.milo.opcua.sdk.server;
 import com.google.common.collect.LinkedHashMultiset;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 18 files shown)


kevinherron:

- [Resolve node identity outside the node manager monitor](https://github.com/eclipse-milo/milo/pull/1875)
- [Preserve replacement nodes during instantiation cleanup](https://github.com/eclipse-milo/milo/pull/1877)
- [Fence type model compilation against concurrent invalidation](https://github.com/eclipse-milo/milo/pull/1881)
- [Preserve type and element boundaries in model fingerprints](https://github.com/eclipse-milo/milo/pull/1884)
- [Isolate nested mutable values in attribute snapshots](https://github.com/eclipse-milo/milo/pull/1887)
- [Resolve overridden declaration references to effective instances](https://github.com/eclipse-milo/milo/pull/1890)
- [Allocate distinct diagnostics element identities after removal](https://github.com/eclipse-milo/milo/pull/1892)

PR #1869, opened 2026-09-05
Preserve Reverse Connect ownership and queued callback delivery

kevinherron:
Preserve connection ownership, callback delivery, and lifecycle order across Reverse Connect cancellation and executor failures.

## Listener binds after shutdown

A startup paused in the bootstrap customizer could bind after shutdown completed. Startup now rechecks the running state under the listener lock immediately before binding.

## Rediscovery after stopping an acceptor

Disconnecting a previously delivered client while its acceptor was stopped left its server key marked active, preventing rediscovery after restart. Delivered-client tracking now continues while the acceptor is stopped.

## Cancelled registration ownership

A cancelled registration could claim a later socket, or cancellation could win after the socket was reserved but before delivery. Exceptional completion now unregisters the selector, and failed claim delivery closes the undelivered connection with a debug diagnostic. Claimed snapshots and counters describe ownership reserved for a selector.

## Channel notification order

Reentrant close from a connected callback or future completion could deliver `[false, true]` to another observer. Serialized transition delivery now preserves `[true, false]`. Milo's observer contract says, "Implementations should keep listener invocation ordered with their own channel lifecycle."

## Executor failures during connect

Rejected claim dispatch could leave connect pending, and rejected handshake-completion dispatch could lose a completed handshake. Claim dispatch failure now closes the socket and fails connect; completed handshakes retain their completion through synchronous fallback.

## Stranded server connection attempts

A rejected listener notification could stop an attempt before transport creation. The shared callback queue now retains delivery and allows the attempt to proceed. Part 6 §7.1.3 says, "Servers shall maintain at least one open socket without an active Session with each Client it is configured to connect to." [Establishing a connection](https://reference.opcfoundation.org/specs/OPC-10000-6/7.1.3)

## Stale attempts after target replacement

An old asynchronous attempt could attach its channel to a replacement target using the same UUID. Callbacks now retain the original registration identity and close channels delivered to a different registration.

## Lost Publish response delivery

Executor rejection after consuming a pending Publish entry could leave its future incomplete. The queue now retains the response callback through initial and continuation dispatch failures. Part 4 §5.14.5.1 describes the supported request flow: "Pipelining involves sending more than one Publish request for each Subscription before receiving a response." [Publish](https://reference.opcfoundation.org/specs/OPC-10000-4/5.14.5.1)

## Queue liveness and serial execution

Dispatch runtime exceptions now run the fallback on the dispatching thread. Finished workers ignore repeated invocations, including when an executor runs a worker before reporting failure, so worker accounting cannot permit overlapping callbacks at concurrency one. Direct executors drain without recursion, and workers submit continuations between bounded batches. Fallback callbacks can run on an I/O or timer thread and must return promptly. These are Milo executor contracts; the OPC UA specification does not prescribe Java cancellation or scheduling behavior.

## Verification

Deterministic regressions reproduced the ownership, dispatch, notification, and serial-execution failures before their fixes. An unfiltered verification run before the final executor guards passed with no failures or errors and three skips. The final state passed formatting, full clean compile, and the selected queue, Publish, and Reverse Connect tests, with no failures, errors, or skips.

### DIFF
--- a/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/reverse/ReverseConnectAcceptor.java
+++ b/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/reverse/ReverseConnectAcceptor.java
@@ -351,11 +351,7 @@ public final class ReverseConnectAcceptor implements AutoCloseable {
 
     ChannelStateObservable.TransitionListener listener =
         connected -> {
-          // After acceptor.stop() the listener is no longer removed from the transport, but it
-          // must not mutate bookkeeping for a stopped acceptor.
-          if (!running.get()) {
-            return;
-          }
+          // Delivered clients outlive stop(); keep ownership current across acceptor restarts.
           if (connected) {
             activeKeys.add(key);
           } else {
--- a/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/reverse/ReverseConnectCandidateSnapshot.java
+++ b/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/reverse/ReverseConnectCandidateSnapshot.java
@@ -38,7 +38,7 @@ import org.jspecify.annotations.Nullable;
  * @param completedAt the time the candidate reached a terminal manager state, or {@code null} while
  *     it can still be matched or claimed.
  * @param rejectionReason the manager-level reason the candidate was rejected, expired, or closed,
- *     or {@code null} for non-terminal and successfully claimed candidates.
+ *     or {@code null} for non-terminal and claimed candidates.
  * @param rejectionStatusCode the TCP error status associated with a rejected or expired candidate,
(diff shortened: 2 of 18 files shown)


kevinherron:

- [Preserve queued callbacks when executor dispatch fails](https://github.com/eclipse-milo/milo/pull/1878)
- [Keep reverse target callbacks with their registration](https://github.com/eclipse-milo/milo/pull/1880)
- [Fence reverse listener binding after shutdown](https://github.com/eclipse-milo/milo/pull/1883)
- [Release cancelled reverse connection claims](https://github.com/eclipse-milo/milo/pull/1886)
- [Restore reverse discovery after disconnecting while stopped](https://github.com/eclipse-milo/milo/pull/1889)
- [Deliver reverse channel transitions in lifecycle order](https://github.com/eclipse-milo/milo/pull/1893)
- [Settle reverse connect after claim dispatch rejection](https://github.com/eclipse-milo/milo/pull/1895)
- [Complete reverse handshakes after executor rejection](https://github.com/eclipse-milo/milo/pull/1898)

PR #1870, opened 2026-09-05
Preserve UASC sequence state and validate Abort chunk recovery

kevinherron:
Preserve the UASC sequence stream after message-size failures and Abort, and release client buffers when partial responses can no longer complete.

## Oversized messages corrupt the sequence stream

Exceeding the peer's chunk-count limit could consume sequence numbers and AEAD nonces for chunks that were never sent. The following fallback response then failed sequence or authentication checks. The encoder now checks the complete message before reserving either value; the fallback and subsequent responses remain decodable under None, CBC, GCM, and ChaCha20-Poly1305.

Part 6 §6.7.2.4 says the sequence number "shall be incremented by exactly one for each MessageChunk sent." [Sequence Header](https://reference.opcfoundation.org/specs/OPC-10000-6/6.7.2.4)

The existing fallback ServiceFault/status choice is preserved. This change does not revise the separate wire error-form/status mapping for oversized responses.

## Abort authentication and recovery

Server handlers previously discarded accumulated chunks on Abort without authenticating it or advancing sequence state. Abort now passes through the existing decoder's security and sequence checks. Invalid signatures, authentication tags, and malformed headers follow the error path.

A valid Abort also remains accepted after the permitted intermediate chunks have arrived, on both server and client. Its size and security checks still apply; the exemption admits at most one additional size-limited chunk before releasing the accumulated message.

Part 6 §6.7.3 says, "The receiver shall check the security on the abort MessageChunk before processing it." It also says the receiver "shall not close the SecureChannel" after a valid Abort. [MessageChunks and error handling](https://reference.opcfoundation.org/specs/OPC-10000-6/6.7.3)

## Client partial chunks survive disconnect

Closing a client channel or removing its handler could leave pooled partial-response buffers retained. Both boundaries now use the same idempotent release-and-clear operation as exception handling. Package documentation explains the buffer transfer and sequence ownership boundaries. Buffer release is a Milo/Netty ownership requirement.

## Verification

Regression tests cover corrupted fallback responses, unauthenticated Abort acceptance, skipped Abort sequence state, retained client buffers, and rejection of valid Abort at the chunk-count boundary. Controls verify valid signed Abort, subsequent requests/responses, and buffer release under None and GCM.

Formatting and full clean compile passed. The selected chunk, UASC, security, and Session tests passed, including ECC Session integration cases, with no failures, errors, or skips. Sensitive regressions failed on their corresponding unchanged implementations. No third-party interoperability run was performed.

### DIFF
--- a/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/channel/ChunkEncoder.java
+++ b/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/channel/ChunkEncoder.java
@@ -278,6 +278,21 @@ public final class ChunkEncoder {
 
       assert (maxPlainTextSize + securityHeaderSize + SECURE_MESSAGE_HEADER_SIZE <= maxChunkSize);
 
+      if (messageBuffer.isReadable() && maxBodySize <= 0) {
+        throw new UaException(
+            StatusCodes.Bad_EncodingLimitsExceeded, "send buffer cannot fit a message body");
+      }
+
+      // No chunk has been emitted yet. Reject the complete message before reserving sequence
+      // numbers or AEAD nonces so a following ServiceFault uses the peer's expected stream state.
+      int remoteMaxChunkCount = parameters.getRemoteMaxChunkCount();
+      if (remoteMaxChunkCount > 0
+          && messageBuffer.readableBytes() > (long) maxBodySize * remoteMaxChunkCount) {
+        throw new UaException(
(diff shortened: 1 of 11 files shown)


kevinherron:

- [Preserve sequence state when chunk limits reject encoding](https://github.com/eclipse-milo/milo/pull/1918)
- [Authenticate Abort chunks and preserve the following message](https://github.com/eclipse-milo/milo/pull/1920)
- [Release partial client responses when handlers terminate](https://github.com/eclipse-milo/milo/pull/1922)

PR #1871, opened 2026-09-05
Fix GDS module compatibility, dictionary registration, and Write results

kevinherron:
This PR fixes GDS module compatibility, dictionary initialization, generated Write error handling, and test-server private-key formats.

## Java module package overlap

Using `milo-sdk-client` and `milo-sdk-client-gds` together on the module path fails because both jars contain `org.eclipse.milo.opcua.sdk.client.gds`. The generated type initializers now live in `org.eclipse.milo.opcua.sdk.client.gds.model`, with registration and documentation updated to match.

## Legacy dictionary registration

Initializing the standard dictionary first skips GDS registration; initializing GDS first puts its entries in the standard dictionary. The regenerated adapter now registers an independent GDS dictionary with unqualified type names. Initialization order is a Milo registration contract.

## Blocking property Write results

Generated blocking writers discard rejected operation statuses and return successfully. The regenerated GDS and namespace-zero writers now throw `UaException` carrying any non-Good result, matching Milo's existing blocking API. OPC UA defines per-operation results; the Java exception mapping is Milo's contract. [Part 4 §5.11.4.2](https://reference.opcfoundation.org/specs/OPC-10000-4/5.11.4.2)

## Test-fixture private key formats

`FakeGdsNamespace` returns raw DER for a PEM request and ignores format/password arguments. It now returns actual PKCS#8 PEM or PKCS#12 PFX, applies the requested password, and rejects unsupported formats. PFX includes the issued certificate and its issuer.

Part 12 §7.9.4 says, "All CertificateManager implementations shall support "PEM" and "PFX"." It also says, "If PFX is used the packet shall include the Certificate and the PrivateKey." [StartNewKeyPairRequest](https://reference.opcfoundation.org/specs/OPC-10000-12/7.9.4)

## Verification

All nine new regression cases fail on the original code at the intended assertions. They resolve the actual modules, initialize both dictionary orders and round-trip a GDS ExtensionObject, parse requested key formats and verify their keys, and exercise rejected and successful Writes through a client/server connection.

The complete selected GDS and TrustList suites passed, including all nine new regressions, with zero failures, errors, or skips. Formatting and full clean compile passed. The [generator changes](https://github.com/kevinherron/opc-ua-codegen2/pull/1) passed both emitted-code runtime regressions, both BACnet generation checks, formatting, and full clean compile. The [regenerated GDS source model](https://github.com/kevinherron/opc-ua-gds-model/pull/1) passed formatting, full clean compile, and verify.

### DIFF
--- a/docs/features/gds-client.md
+++ b/docs/features/gds-client.md
@@ -31,13 +31,14 @@ model does. It lives in subpackages of the existing modules:
 | `milo-stack-core` | `org.eclipse.milo.opcua.stack.core.gds`             | `GdsNodeIds` (one `ExpandedNodeId` per GDS node), `DataTypeInitializer`                          |
 | `milo-stack-core` | `org.eclipse.milo.opcua.stack.core.gds.types`       | `ApplicationRecordDataType` with its codec and `StructureDefinition`                             |
 | `milo-dtd-core`   | `org.eclipse.milo.opcua.sdk.core.dtd.gds`           | `BinaryDataTypeDictionaryInitializer` for the legacy type dictionary mechanism (deprecated)      |
-| `milo-sdk-client` | `org.eclipse.milo.opcua.sdk.client.gds`             | `ObjectTypeInitializer`, `VariableTypeInitializer`                                                |
+| `milo-sdk-client` | `org.eclipse.milo.opcua.sdk.client.gds.model`             | `ObjectTypeInitializer`, `VariableTypeInitializer`                                                |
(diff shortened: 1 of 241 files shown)


kevinherron:

- [Keep GDS client jars usable together as Java modules](https://github.com/eclipse-milo/milo/pull/1915)
- [Report rejected property Writes to blocking callers](https://github.com/eclipse-milo/milo/pull/1917)
- [Register the legacy GDS dictionary independently of namespace zero](https://github.com/eclipse-milo/milo/pull/1919)
- [Honor the requested private-key format in the GDS test server](https://github.com/eclipse-milo/milo/pull/1921)

PR #1872, opened 2026-09-05
Fix alias deletion, shutdown ordering, and change versions

kevinherron:
This PR fixes five alias-management issues involving deletion, lifecycle ordering, and change-version tracking.

### Remove alias references across node managers

Deleting a target or a whole alias could report success while leaving references in another NodeManager. Recreating the same alias name could then restore a deleted target or category membership. Reference removal now covers every registered manager that can store the association. Whole-alias deletion preserves deletion of its owned children before removing the saved cross-manager references.

### Prepare all category versions before deleting an entry

A version-store failure could leave one deletion entry partly applied. Each entry now prepares the union of affected category versions before removing its first target.

[Part 17 §6.3.5](https://reference.opcfoundation.org/specs/OPC-10000-17/6.3.5) requires: "If all targets for an AliasNames array entry cannot be deleted, then none of the targets are deleted." Preparing versions first prevents a persistence failure from deleting only some aliases in that entry. Existing custom NodeManager failure behavior is unchanged.

### Detach surviving child categories

Removing a parent category could leave its surviving children attached to a deleted parent. Category removal now detaches surviving child categories as well as aliases.

### Reject mutations after shutdown begins

A caller could pass the lifecycle check, wait for the mutation lock, and then change the address space after shutdown. Public mutations now check lifecycle state while holding that lock.

### Preserve change versions when a category ID is reused

Recreating a category NodeId could reuse an observed LastChange for different contents. The manager now retains the category's version high-water mark across removal and recreation.

[Part 17 §6.3.1](https://reference.opcfoundation.org/specs/OPC-10000-17/6.3.1) states: "The LastChange shall be persisted." The reuse fix preserves Milo's version-store contract within the manager's lifetime. Preserving removed-category versions across manager restarts still requires the store to retain their entries.

### Verification

Formatting, a full clean compile, and the selected alias suites passed with no failures, errors, or skips. The original 12 regressions fail at their intended assertions without the fixes, including real wire deletion and deterministic shutdown interleavings.

Three additional regressions fail before the whole-alias cleanup correction and pass afterward: public and wire deletion followed by same-name recreation, and deletion with category membership stored in another manager. A fourth case confirms that owned-child deletion still works. The unchanged in-memory version-store tests passed in the earlier verification run.

### DIFF
--- a/opc-ua-sdk/integration-tests/src/test/java/org/eclipse/milo/opcua/sdk/server/aliases/AliasLifecycleConsistencyTest.java
+++ b/opc-ua-sdk/integration-tests/src/test/java/org/eclipse/milo/opcua/sdk/server/aliases/AliasLifecycleConsistencyTest.java
@@ -1,0 +1,400 @@
+/*
+ * Copyright (c) 2026 the Eclipse Milo Authors
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.milo.opcua.sdk.server.aliases;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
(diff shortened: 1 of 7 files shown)


kevinherron:

- [Remove alias associations from every registered node manager](https://github.com/eclipse-milo/milo/pull/1896)
- [Detach surviving child categories when their parent is removed](https://github.com/eclipse-milo/milo/pull/1899)
- [Prepare every category version before deleting one alias entry](https://github.com/eclipse-milo/milo/pull/1902)
- [Reject alias mutations that resume after shutdown](https://github.com/eclipse-milo/milo/pull/1905)
- [Preserve category versions when their node identifiers are reused](https://github.com/eclipse-milo/milo/pull/1908)

PR #1873, opened 2026-09-05
Fix Condition event consistency and shelving Method ownership

kevinherron:
This PR fixes three Condition lifecycle issues involving event consistency and shelving Method dispatch.

### Keep shelving state consistent within one event

Lazy shelving expiry could change a Condition's EventId and state while its fields were being selected for one notification or copied for refresh. Expiry now resolves before those operations, and field reads cannot trigger another shelving transition partway through them. Ordinary UnshelveTime reads retain their lazy-expiry fallback.

[Part 5 §6.4.2](https://reference.opcfoundation.org/specs/OPC-10000-5/6.4.2) defines the identity contract: "EventId is generated by the Server to uniquely identify a particular Event Notification." The fix keeps that identity and its shelving state consistent across the notification's fields.

### Retire Method handlers when a Condition is unregistered

Copied Method nodes could retain handlers belonging to a retired Condition behavior. Unregistering a Condition now removes handlers it installed only if it still owns them. Application handlers and handlers installed by a replacement behavior survive. Calls that already obtained a handler may finish against the retired behavior.

### Dispatch shared shelving methods through either supported ObjectId

Shared shelving methods previously failed to find their behavior through a nested ShelvingState ObjectId. Dispatch now resolves both that target and the ConditionId, while leaving the shared type Method's handler unchanged. The nested route accepts only Methods owned by the shelving state machine, so an Acknowledge call still requires the ConditionId.

[Part 9 §5.8.17.4](https://reference.opcfoundation.org/specs/OPC-10000-9/5.8.17.4) defines the Shelving object's NodeId as the normal call target and also requires that "all Servers shall also allow Clients to call the TimedShelve Method by specifying ConditionId as the ObjectId".

### Verification

Formatting, a full clean compile, and the selected Condition and event-handling suites passed with no failures, errors, or skips. Regressions exercise coherent event selection and refresh, wire calls after unregister, preserved replacement handlers, and both shelving call targets. The base implementation failed four of the first six new cases; the remaining two provide compatibility controls.

A further wire regression rejects Acknowledge through the nested ShelvingState ObjectId and then proves that the same Method and arguments succeed through the ConditionId. It failed before the dispatch ownership restriction and passes with the final implementation.

### DIFF
--- a/opc-ua-sdk/integration-tests/src/test/java/org/eclipse/milo/opcua/sdk/server/conditions/ConditionDispatchLifecycleTest.java
+++ b/opc-ua-sdk/integration-tests/src/test/java/org/eclipse/milo/opcua/sdk/server/conditions/ConditionDispatchLifecycleTest.java
@@ -1,0 +1,277 @@
+/*
+ * Copyright (c) 2026 the Eclipse Milo Authors
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.milo.opcua.sdk.server.conditions;
+
+import static java.util.Objects.requireNonNull;
+import static org.eclipse.milo.opcua.stack.core.types.builtin.unsigned.Unsigned.ushort;
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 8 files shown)


kevinherron:

- [Resolve shelving expiry before selecting Condition event fields](https://github.com/eclipse-milo/milo/pull/1906)
- [Retire copied Method handlers when their Condition is unregistered](https://github.com/eclipse-milo/milo/pull/1909)
- [Resolve shared shelving Methods through their nested Object](https://github.com/eclipse-milo/milo/pull/1912)

PR #1874, opened 2026-09-05
Preserve failed activation state and select compatible X509 policies

kevinherron:
This PR repairs session failure cleanup and X509 authentication policy selection.

### Close unpublished sessions

Failed CreateSession header negotiation could leave a provisional session timer alive. The server now closes the unpublished session on failure and cancels its timeout. This follows Milo's resource-ownership contract.

### Preserve state when activation fails

Failed ActivateSession header negotiation could partially commit identity, nonce, locale, channel ownership, or active-session registration. The server now prepares the additional response header before changing that state, so a failed activation preserves the previously usable session. A secured-channel retry can still sign the nonce from the last successful response.

[Part 4 §5.7.3.1](https://reference.opcfoundation.org/specs/OPC-10000-4/5.7.3.1) states: "If no errors occur the Server replaces the user identity for the Session."

### Select an advertised policy compatible with the user certificate

X509 identity selection now checks the user's actual public key and EC curve while scanning advertised Certificate policies. It skips policies resolving to None and selects a compatible signing policy before requesting the private key. [Part 4 §7.40.5](https://reference.opcfoundation.org/specs/OPC-10000-4/7.40.5) requires that "the Server shall provide a distinct UserTokenPolicy for each CertificateKeyAlgorithm supported."

### Allow certificate tokens to use a different key family from the channel

Client and server also applied a key-family restriction to certificate tokens that belongs only to username and issued tokens. [Part 4 §7.41](https://reference.opcfoundation.org/specs/OPC-10000-4/7.41) says: "If the tokenType is CERTIFICATE, the securityPolicyUri may be any valid SecurityPolicy." RSA user certificates can therefore authenticate over an ECC SecureChannel. The username/issued-token key-family rule no longer rejects this valid combination. Existing runtime limitations for enhanced certificate-token signatures over legacy secured channels remain documented and tested.

### Verification

Verification passed: formatting, full clean compile, and selected session-header, endpoint-binding, X509, user-token, and certificate-compatibility tests. Detached-base regressions reproduce the timer leak, all three activation failure paths, incompatible policy selection, and incorrect cross-algorithm rejection. Additional regressions verify signed retries with the last delivered nonce on the same or a replacement channel, and skip explicit or inherited None before a usable certificate signing policy. Both secured retry cases also passed after adding an assertion that the original channel remains usable immediately after header rejection.

### DIFF
--- a/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/identity/X509IdentityProvider.java
+++ b/opc-ua-sdk/sdk-client/src/main/java/org/eclipse/milo/opcua/sdk/client/identity/X509IdentityProvider.java
@@ -19,13 +19,12 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Supplier;
-import java.util.stream.Stream;
 import org.eclipse.milo.opcua.stack.core.StatusCodes;
 import org.eclipse.milo.opcua.stack.core.UaException;
+import org.eclipse.milo.opcua.stack.core.security.CertificateCompatibility;
 import org.eclipse.milo.opcua.stack.core.security.ChannelBoundSignatureData;
 import org.eclipse.milo.opcua.stack.core.security.SecurityPolicy;
 import org.eclipse.milo.opcua.stack.core.security.SecurityPolicyProfile;
-import org.eclipse.milo.opcua.stack.core.security.UserTokenSecurityPolicyRules;
 import org.eclipse.milo.opcua.stack.core.types.builtin.ByteString;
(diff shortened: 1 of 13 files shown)


kevinherron:

- [Close provisional Sessions when response preparation fails](https://github.com/eclipse-milo/milo/pull/1894)
- [Prepare activation responses before changing Session state](https://github.com/eclipse-milo/milo/pull/1897)
- [Select a user token policy compatible with the certificate](https://github.com/eclipse-milo/milo/pull/1900)
- [Allow certificate tokens to use a different key family from the channel](https://github.com/eclipse-milo/milo/pull/1903)

PR #1966, opened 2026-09-09
Encode null elements in structured Variant arrays

kevinherron:
Variant encoding throws a NullPointerException when a typed structure array contains a null element, such as `new XVType[] {new XVType(1.0, 2.0f), null}`. Preserve these positions through binary, JSON, and XML encoding without passing null to a structure codec. Binary writes the null ExtensionObject form, JSON writes literal `null` in both compact and verbose modes, and XML writes `xsi:nil` elements. XML distinguishes array elements from optional scalar fields so null positions are retained.

The regressions round-trip concrete and UaStructuredType arrays and supported 2D/3D Matrices, including empty arrays and all-null and mixed-null elements. They check wire null representations, decoded values, element order, and dimensions. Positive tests reproduced the failures before each repair.

Empty Matrices are excluded from the new wire tests because binary Variant dimensions must all be positive. The existing encoder behavior is tracked separately in #1968. Adjacent enum and OptionSet handling remains unchanged.

Independent review found no remaining issues. Verification passed on the final source with Java 17: `mise exec -- mvn -q spotless:apply`, `mise exec -- mvn -q -pl opc-ua-stack/encoding-json,opc-ua-stack/encoding-xml -am clean verify`, `mise exec -- mvn -q clean compile`, and `mise exec -- mvn -q clean verify`. The affected-module reactor passed 2,107 tests. Full reactor verification passed 4,841 tests with zero failures or errors and four existing skips. All 23 binary, 46 JSON and 24 XML regression cases passed.

### DIFF
--- a/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/encoding/binary/OpcUaBinaryEncoder.java
+++ b/opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/encoding/binary/OpcUaBinaryEncoder.java
@@ -686,7 +686,8 @@ public class OpcUaBinaryEncoder implements UaEncoder {
     if (structure) {
       UaStructuredType struct = (UaStructuredType) value;
 
-      ExtensionObject extensionObject = ExtensionObject.encode(context, struct);
+      ExtensionObject extensionObject =
+          struct != null ? ExtensionObject.encode(context, struct) : null;
 
       encodeBuiltinType(typeId, extensionObject);
     } else if (enumeration) {
--- a/opc-ua-stack/stack-core/src/test/java/org/eclipse/milo/opcua/stack/core/encoding/binary/StructuredVariantSerializationTest.java
+++ b/opc-ua-stack/stack-core/src/test/java/org/eclipse/milo/opcua/stack/core/encoding/binary/StructuredVariantSerializationTest.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2026 the Eclipse Milo Authors
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.milo.opcua.stack.core.encoding.binary;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import java.util.stream.Stream;
+import org.eclipse.milo.opcua.stack.core.OpcUaDataType;
+import org.eclipse.milo.opcua.stack.core.encoding.DefaultEncodingContext;
+import org.eclipse.milo.opcua.stack.core.types.UaStructuredType;
+import org.eclipse.milo.opcua.stack.core.types.builtin.ExtensionObject;
+import org.eclipse.milo.opcua.stack.core.types.builtin.Matrix;
+import org.eclipse.milo.opcua.stack.core.types.builtin.NodeId;
+import org.eclipse.milo.opcua.stack.core.types.builtin.Variant;
+import org.eclipse.milo.opcua.stack.core.types.structured.XVType;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class StructuredVariantSerializationTest {
+
+  private final ByteBuf buffer = Unpooled.buffer();
+  private final OpcUaBinaryEncoder encoder =
+      new OpcUaBinaryEncoder(DefaultEncodingContext.INSTANCE).setBuffer(buffer);
+  private final OpcUaBinaryDecoder decoder =
+      new OpcUaBinaryDecoder(DefaultEncodingContext.INSTANCE).setBuffer(buffer);
+
+  @AfterEach
+  void releaseBuffer() {
+    buffer.release();
+  }
+
+  // Typed structure arrays must preserve null positions as null ExtensionObjects on the wire.
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("structureArrays")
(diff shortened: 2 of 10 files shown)


kevinherron:
A Variant containing a typed structure array failed when encoding reached a null element. The binary, JSON and XML converters tried to pass that null to a structure codec. XML also omitted null ExtensionObject array elements instead of preserving their positions.

Null elements now use the binary null ExtensionObject form, JSON null in both modes, and XML nil elements. Non-null structures still use their registered codecs. The same handling applies to supported Matrices.

The positive tests reproduced the failures before repair and now round-trip concrete and UaStructuredType arrays and Matrices with all-null and mixed-null elements. Empty arrays remain covered. New wire Matrix tests use positive dimensions; the existing empty-Matrix encoding behavior is tracked in #1968. The binary documentation cites Part 6 section 5.2.2.15, and the XML tests check every nil position while retaining scalar-field omission.

Independent review findings are resolved. Formatting, clean compilation and clean affected-module and full-reactor verification passed. All 23 binary, 46 JSON and 24 XML regression cases pass; the full reactor ran 4,841 tests with zero failures or errors and four existing skips.

opc-ua-stack/stack-core/src/test/java/org/eclipse/milo/opcua/stack/core/encoding/binary/StructuredVariantSerializationTest.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  static Stream<Arguments> structureMatrices() {
+    return structureArrays()
+        .flatMap(
+            arguments -> {
+              Object[] values = arguments.get();
+              String name = (String) values[0];
+              UaStructuredType[] elements = (UaStructuredType[]) values[1];
+              int rows = elements.length / 2;
```
kevinherron, line 100: The empty cases produce `rows == 0`, so this builds `Matrix(new XVType[0], {0, 2})` and `{1, 0, 2}`, and the test then asserts a `0xD6` mask. That writes an ArrayDimensions field containing a zero-length dimension.

Part 6 Table 26 (§5.2.2.16) says the ArrayDimensions field "shall only be present if the number of dimensions is 2 or greater **and all dimensions have a length greater than 0**," and that all dimensions "shall be specified and shall be greater than zero." A decoder is entitled to raise `Bad_DecodingError` on this.

This is pre-existing Matrix behavior, not something this PR introduces, but the test locks it in. Either drop the empty Matrix cases and file the encoder issue separately, or fix the encoder to fall back to the plain array mask when any dimension is <= 0.

opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/encoding/binary/OpcUaBinaryEncoder.java
```diff
@@ -686,7 +686,8 @@ private void encodeValue(
     if (structure) {
       UaStructuredType struct = (UaStructuredType) value;
 
-      ExtensionObject extensionObject = ExtensionObject.encode(context, struct);
+      ExtensionObject extensionObject =
+          struct != null ? ExtensionObject.encode(context, struct) : null;
```
kevinherron, line 690: This is the right call per §5.2.2.15, but it's the only one of the three encoders that got it. `OpcUaJsonEncoder:971`/`:1578` and `OpcUaXmlEncoder:711`/`:1941` still cast an unchecked null into `ExtensionObject.encode`.

kevinherron (commented):
## Part 6 compliance

§5.2.6 says:

> Structures do not have a null value. If an encoder is written in a programming language that allows structures to have null values, then the encoder shall create a new instance with DefaultValues for all fields and serialize that. Encoders shall not generate an encoding error in this situation.

That rule governs the Structure encoding itself, where a structure is a bare sequence of fields with nowhere to put a null marker. It doesn't reach this case. A structure inside a Variant is carried as an ExtensionObject, and ExtensionObject has a defined null:

- Table 1: ExtensionObject, Nullable = Yes.
- §5.2.2.15: "A null value for an ExtensionObject is encoded with a TypeId set to 'i=0' and the Encoding set to 0."
- §5.2.2.16 puts no constraint on Variant array elements beyond encoding each according to its built-in type.

So the bytes are well-formed and any conformant stack can decode them. The approach is right. A strict reading of §5.2.6 would have the encoder substitute a default-constructed instance, but that isn't possible for `UaStructuredType[]` (the element type isn't recoverable from the array), and it would invent data the caller didn't write. The pre-PR NPE is the one behavior §5.2.6 unambiguously prohibits.

Two things to fix before this lands.

### 1. The fix is binary-only

JSON and XML have the identical unguarded cast on the Variant structure path and will NPE the same way:

- `opc-ua-stack/encoding-json/.../OpcUaJsonEncoder.java:971` (scalar and array element) and `:1578` (Matrix)
- `opc-ua-stack/encoding-xml/.../OpcUaXmlEncoder.java:711` (array) and `:1941` (Matrix)

If null-tolerance is now the contract for structures in Variants, it should hold across all three encodings, with round-trip coverage for the null forms (XML `nil` per §5.3.5, JSON null).

### 2. The new Matrix test pins non-conformant dimensions

See the inline comment on `structureMatrices()`.

### Minor

The package-info paragraph on null structure elements is the non-obvious part of the encoder's behavior. Worth citing §5.2.2.15 there so the next reader doesn't have to re-derive why it isn't a §5.2.6 violation.

opc-ua-stack/stack-core/src/test/java/org/eclipse/milo/opcua/stack/core/encoding/binary/StructuredVariantSerializationTest.java
kevinherron: Removed the empty Matrix wire cases while retaining empty typed arrays and positive-dimension 2D/3D Matrices. Filed #1968 for the existing binary encoder behavior, with the actual bytes and the plain-array encoding required by Table 26.

opc-ua-stack/stack-core/src/main/java/org/eclipse/milo/opcua/stack/core/encoding/binary/OpcUaBinaryEncoder.java
kevinherron: Added the corresponding JSON and XML repairs. JSON writes literal null for structure elements in both encoding modes. XML guards structure conversion and distinguishes array/Matrix elements from optional scalar fields so nil positions are retained. Positive round-trip tests cover concrete and UaStructuredType arrays and supported Matrices, including empty arrays, all-null and mixed-null cases. The binary package documentation now cites Part 6 section 5.2.2.15. The clean affected-module and full-reactor gates pass, including all 23 binary, 46 JSON and 24 XML regression cases.

kevinherron:
@greptileai Please review the updated commit befb2a3eaa6e854511a41fb62571374a224431a3, including the changes addressing the previous feedback.

PR #1969, opened 2026-09-10
Omit binary Variant dimensions for empty Matrices

### LINKED ISSUE #1968: Omit binary Variant dimensions for empty Matrices
`OpcUaBinaryEncoder.encodeVariant` writes the ArrayDimensions flag and dimension values for every non-null Matrix, including Matrices with a zero-length dimension. This produces a nonconforming binary Variant that a stricter peer may reject.

The behavior is present on `integration/1.2` at `952b61f9b17a22a8b9f7dfd4145699720483519d` (`1.2.0-SNAPSHOT`). An empty typed structure Matrix demonstrates it without encoding any structure elements:

```java
import io.netty.buffer.ByteBuf;
import io.netty.buffer.ByteBufUtil;
import io.netty.buffer.Unpooled;
import org.eclipse.milo.opcua.stack.core.encoding.DefaultEncodingContext;
import org.eclipse.milo.opcua.stack.core.encoding.binary.OpcUaBinaryEncoder;
import org.eclipse.milo.opcua.stack.core.types.builtin.Matrix;
import org.eclipse.milo.opcua.stack.core.types.builtin.Variant;
import org.eclipse.milo.opcua.stack.core.types.structured.XVType;

ByteBuf buffer = Unpooled.buffer();
try {
  var matrix = new Matrix(new XVType[0], new int[] {0, 2});
  new OpcUaBinaryEncoder(DefaultEncodingContext.INSTANCE)
      .setBuffer(buffer)
      .encodeVariant(new Variant(matrix));
  System.out.println(ByteBufUtil.hexDump(buffer));
} finally {
  buffer.release();
}
```

The encoder writes `d600000000020000000000000002000000`: ExtensionObject array mask `0xD6`, zero elements, and dimensions `[0, 2]`. Milo's own decoder accepts this, so a local round-trip test does not establish conformance.

[OPC UA Part 6 §5.2.2.16, Table 26](https://reference.opcfoundation.org/Core/Part6/v105/docs/5.2.2.16) permits ArrayDimensions only for at least two dimensions whose lengths are all positive. It also specifies an ArrayLength of zero when any dimension is nonpositive. This value should therefore use the plain ExtensionObject array mask `0x96`, ArrayLength zero and no dimensions (`9600000000`). The same issue affects other element types.

Update the binary Variant encoder to omit dimensions and write an empty array for nonpositive dimension lengths. Cover zero and negative lengths in different dimension positions, preserve the built-in element type, and retain dimensions for valid nonempty Matrices. Keep this separate from nullable structure-element handling because it changes the wire representation and the shape available after decoding.

[PR #1791](https://github.com/eclipse-milo/milo/pull/1791) preserved type metadata for empty Matrices. This report concerns the binary Variant's ArrayDimensions field, so it does not require removing that local representation or its metadata.

kevinherron:
Fixes #1968.

When a Variant holds a Matrix with a zero or negative dimension, the binary encoder wrote the ArrayDimensions bit and the dimension values anyway. Part 6 §5.2.2.16 (Table 26) says ArrayDimensions may only appear when there are at least two dimensions and every dimension is greater than 0, and that ArrayLength is 0 when any dimension is not. A stricter peer can reject the old form. Milo's own decoder accepts it, so a local round trip did not catch this.

The encoder now writes such a Matrix as a plain empty array of its built-in element type: the array bit only, ArrayLength 0, and no dimensions. For an empty `XVType` Matrix with dimensions `[0, 2]`, the bytes change like this:

```diff
-d6 00000000 02000000 00000000 02000000   mask=ExtensionObject|Array|Dims, length 0, dims [0, 2]
+96 00000000                              mask=ExtensionObject|Array,      length 0
```

Nonempty Matrices are unchanged and still carry their dimensions. The Matrix constructors only assert consistency, so a Matrix can carry elements alongside a dimension of 0; the encoder now rejects that with Bad_EncodingError rather than writing an empty array and silently dropping the elements. The element type is preserved, so an empty enumeration Matrix still encodes as Int32 and an empty option set Matrix as its backing unsigned type. The Matrix class itself and the type metadata it keeps for empty values (from #1791) are not touched.

One consequence: the receiving side can no longer tell an empty Matrix from an empty one-dimensional array, because the wire form is the same. The server's Method argument shape check compares the decoded rank to the declared ValueRank, so an empty Matrix sent to a ValueRank 2 argument now decodes as rank 1 and is rejected with Bad_TypeMismatch. Two integration test cases relied on the old encoding to pass that check, and this PR removes them; local (non-wire) cases for the same values remain. Whether a ValueRank ≥ 2 argument should accept an empty array from the wire is a server validation question, left for a follow-up.

Verification: `spotless:apply`, `clean compile`, the stack-core binary encoder and decoder test classes, and `MethodArgumentValidationTest` in integration-tests (with `-am`) all pass locally. New golden-byte tests cover zero and negative lengths in first, middle, and last dimension positions, structure, enumeration, option set, boxed, and primitive element types, and nonempty controls that keep their dimensions.

### DIFF
--- a/opc-ua-sdk/integration-tests/src/test/java/org/eclipse/milo/opcua/sdk/server/methods/MethodArgumentValidationTest.java
+++ b/opc-ua-sdk/integration-tests/src/test/java/org/eclipse/milo/opcua/sdk/server/methods/MethodArgumentValidationTest.java
@@ -173,7 +173,8 @@ class MethodArgumentValidationTest extends AbstractClientServerTest {
                 baseValue(1, new UaStructuredType[] {null, xv}),
                 baseValue(2, new Matrix(new XVType[] {null, xv}, new int[] {1, 2})),
                 baseValue(2, new Matrix(new UaStructuredType[] {null, xv}, new int[] {1, 2})),
-                // Local shape validation does not establish a wire encoding for empty Matrices.
+                // Local only: Part 6 5.2.2.16 encodes an empty Matrix as an empty array with no
+                // dimensions, so the wire form arrives as a one-dimensional empty array.
                 baseValue(2, new Matrix(new String[0], new int[] {0, 2})),
                 baseValue(2, new Matrix(new XVType[0], new int[] {0, 2})),
                 baseValue(-1, Matrix.ofNull())))
@@ -368,7 +369,6 @@ class MethodArgumentValidationTest extends AbstractClientServerTest {
         shape(2, new Matrix(new Integer[] {1, 2}, new int[] {1, 2}), true),
         shape(3, new Matrix(new Integer[] {1, 2}, new int[] {1, 2}), false),
         shape(3, new Matrix(new Integer[] {1, 2}, new int[] {1, 1, 2}), true),
-        shape(2, new Matrix(new Integer[0], new int[] {0, 2}), true),
         shape(-1, null, true),
         shape(1, null, true),
         shape(2, null, true),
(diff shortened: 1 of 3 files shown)


kevinherron:
@greptileai review

kevinherron:
@greptileai review

PR #2004, opened 2026-09-15
Preserve null array fields in VERBOSE JSON

### LINKED ISSUE #2003: VERBOSE JSON drops null array fields
`OpcUaJsonEncoder.encodeArray` returns immediately when the array is null, regardless of the selected encoding mode. This drops mandatory null array fields in VERBOSE JSON. It also makes a present-null optional array disappear, losing presence information once VERBOSE correctly omits the encoding mask.

A generic codec that invokes:

```java
encoder.encodeInt32Array("Numbers", null);
encoder.encodeEnumArray("Enums", null);
encoder.encodeStructArray("Structures", null, NodeIds.Range);
encoder.encodeExtensionObjectArray("Objects", null);
encoder.encodeInt32("Tail", 88);
```

produces `{"Tail":88}` in VERBOSE mode. Expected:

```json
{"Numbers":null,"Enums":null,"Structures":null,"Objects":null,"Tail":88}
```

The same helper correctly produces empty JSON arrays for empty Java arrays. The missing null handling is shared by the direct builtin, enum, structure, and ExtensionObject array paths.

[Part 6 5.4.6, Table 45](https://reference.opcfoundation.org/specs/OPC-10000-6/5.4.6) requires null structure fields to be JSON null in VERBOSE and omitted in COMPACT. [Part 6 5.4.7, Table 47](https://reference.opcfoundation.org/specs/OPC-10000-6/5.4.7) makes the optional distinction explicit: absent stays omitted; present-null becomes a named JSON null in VERBOSE. This is an encoder issue, separate from the closed omitted-structure decoder issue #1773.

A minimal reproducer (JDK 17+ and Milo core/JSON on the classpath):

```java
import org.eclipse.milo.opcua.stack.core.encoding.*;
import org.eclipse.milo.opcua.stack.core.encoding.json.OpcUaJsonEncoder;
import org.eclipse.milo.opcua.stack.core.types.structured.Structure;

class Repro {
  public static void main(String[] args) throws Exception {
    var context = new DefaultEncodingContext();
    var codec = new GenericDataTypeCodec<Structure>() {
      public Class<Structure> getType() { return Structure.class; }
      public Structure decodeType(EncodingContext c, UaDecoder d) {
        throw new UnsupportedOperationException();
      }
      public void encodeType(EncodingContext c, UaEncoder e, Structure value) {
        e.encodeInt32Array("Numbers", null);
        e.encodeInt32("Tail", 88);
      }
    };
    try (var encoder = new OpcUaJsonEncoder(context)) {
      encoder.setEncoding(OpcUaJsonEncoder.Encoding.VERBOSE);
      encoder.encodeStruct(null, new Structure() {}, codec);
      System.out.println(encoder.getOutputString());
    }
  }
}
```

Actual `{"Tail":88}`; expected `{"Numbers":null,"Tail":88}`. For an optional field, a codec with a set presence bit calls `encodeInt32Array("Numbers", null)` and receives the same omission. A codec whose bit is clear does not call the field encoder and should continue omitting the member.

Observed on published core `1.2.0-20260915.205718-70` and JSON `1.2.0-20260915.205718-69`. The JSON JAR SHA-256 is `627f97e44c7568a04d262cc0092ed800273f7f4812b1af404a361068a62a5131`; the relevant source is unchanged at `c757feb704cbd26028f7b86e3dfbc9d4fa8a37ce`. The pending deployment at that revision contains only an unrelated Struct.Member hashing change, so there is no already-merged null-array repair awaiting publication.

The shared helper should emit named JSON null for VERBOSE structure fields, preserve COMPACT omission, and retain positional nulls where needed. Tests should distinguish absent, present-null, empty, and populated array fields in both modes.

kevinherron:
When a structure codec encodes a null array, the JSON encoder currently drops the field in both modes. In VERBOSE this removes mandatory fields and makes a present-null optional field indistinguishable from an absent field. The shared array helper now emits JSON null in VERBOSE and in built-in contexts, while preserving COMPACT structure-field omission.

For a codec that writes a null `Numbers` array followed by `Tail = 88`:

| Context | Before | After |
| --- | --- | --- |
| VERBOSE structure | `{"Tail":88}` | `{"Numbers":null,"Tail":88}` |
| COMPACT structure | `{"Tail":88}` | `{"Tail":88}` |
| Positional array value, either mode | Value disappears | `null` occupies its position |

This applies to built-in, enum, structure (both type-ID overloads), and ExtensionObject arrays. Codecs still control optional presence by skipping absent fields. Empty and populated arrays retain their existing encoding. The rules follow [Part 6 §5.4.6](https://reference.opcfoundation.org/specs/OPC-10000-6/5.4.6) and [§5.4.7](https://reference.opcfoundation.org/specs/OPC-10000-6/5.4.7). Both array decoder paths now consume explicit JSON null and return a null Java array, so emitted null values can be read without failing or disturbing the following field. Encoding-mask handling and general VERBOSE decoding support are outside this change.

## Verification

Regression tests reproduced the original encoder omission and the decoder rejection of explicit null arrays before their respective fixes. Tests assert JSON output and decode it back, covering omitted and explicit null fields, empty arrays, populated COMPACT arrays, and positional null arrays. Each decode checks a following value to enforce reader position. VERBOSE checks cover null and empty arrays plus Int32 values; they do not establish general VERBOSE round-trip support. Independent subagent reviews of both commits found no issues.

Passed:

- `mise exec -- mvn -q spotless:apply`
- `mise exec -- mvn -q clean compile`
- `mise exec -- mvn -q -pl opc-ua-stack/encoding-json -am test -Dsurefire.failIfNoSpecifiedTests=false -Dtest='*Json*Test'` (382 JSON tests, no failures, errors, or skips)

The full repository test suite was not run.

Fixes #2003.


kevinherron:
@greptile Please re-review the latest commit, 92ba5716c. Both array decoder paths now consume explicit JSON null values. Regression tests now decode the encoder output and verify null, empty, populated COMPACT, and positional arrays, including the following field/value. Formatting, clean compilation, and 382 JSON tests passed.

